PR #89, opened 2015-01-15
Preparing the environment

JensTimmerman:

PR #92, opened 2015-01-16
Brussels changes in CH2

hajgato:


intro-HPC/ch_account.tex
```diff
@@ -382,6 +406,9 @@ \section{First Time connection to the \hpc}
       \ifgent
       Make sure the fingerprint in the Alert matches \strong{2f:0c:f7:76:87:57:f7:5d:2d:7b:d1:a1:e1:86:19:f3}
       \fi
+      \ifbrussel
+      Make sure the fingerprint in the Alert matches \strong{41:68:9b:c2:f0:1c:d3:62:14:b6:71:f0:36:8a:7d:ec}
```
JensTimmerman: I created a new macro for this `\fingerprint` in https://github.com/hpcugent/vsc_user_docs/pull/89

intro-HPC/ch_account.tex
```diff
@@ -267,6 +283,14 @@ \subsection{Users of the \association}
 to share your Mail and Affiliation with KU Leuven (which actually means the VSC accountpage)
 Select the 'Remember' checkbox and click 'Yes, continue'.
 \fi
+
+\ifbrussel
+ \includegraphics*[width=5.73in, height=4.00in, keepaspectratio=false]{ch2-browser-authenticate-brussel}
```
JensTimmerman: thank you for the screenshots, we should probably have a system where we can import images from their institute subdirectories? A latex macro could be created for this?
so we could put `\includeinstitutegraphics*[width=5.73in, height=4.00in, keepaspectratio=false]{ch2-browser-authenticate}` which looks into the corresponding 'institute' directory and takes the correct graphic?

intro-HPC/ch_account.tex
hajgato: I was not aware of that.... What should I do with it? Delete this part?

intro-HPC/ch_account.tex
hajgato: Indeed we should find a solution for this. In theory in many cases we could include an isnstitute specific graphic. (unfortunately I do not know how to do that...)

intro-HPC/ch_account.tex
hajgato: As well as I was not aware with many other changes. Maybe I have to delete this PR.

JensTimmerman:
I can merge this in to get your images, and then fix the conflicts in my pull request.

boegel:
@JensTimmerman: feel free to merge if you think this is good to go, since you're on top of all of this right now

PR #100, opened 2015-08-20
Kl edits 20150818

klust:
I gathered all corrections I made over the past months and did not yet send to the main repository in previous requests.


boegel:
@JensTimmerman can you take a look at this?

intro-HPC/ch_running_jobs_with_input.tex
```diff
@@ -540,7 +578,7 @@ \subsection{Check your quota}
 3.5M %\homedir/primes.txt%
 \end{prompt}
 
-We also want to mention the \emph{tree} command, as it also provides an easy
+We also want to mention the \texttt{tree} command, as it also provides an easy
```
JensTimmerman: we should probably also mention you shouldn't really use this command all that often, like, not in job scripts etc. since it can put a heavy strain on the file system. Same with `du` actually.

style.tex
```diff
@@ -59,6 +59,7 @@
 \lstset{basicstyle=\ttfamily,breaklines=true, keepspaces=true}
 \lstdefinestyle{prompt}{
 % language=bash,
+  basicstyle=\footnotesize\ttfamily,
```
JensTimmerman: addition of these basicstyle seem to break the minipage setup declared in `\lstnewenvironment{prompt}`
prompts and flattext are now again split over multiple pages.

JensTimmerman:
Thank you!

PR #101, opened 2015-09-24
update Oct 2015

sbecuwe:
- fixed typos
- updated vscentrum urls
- changed tabular header for full page width tables
- make Glossary appear automatically in TOC instead of using addcontentsline


JensTimmerman:
looks good, thank you for these fixes!

PR #110, opened 2016-02-19
Revert epauwels merge

itkovian:
Merge was made without addressing the remarks
- still contained typos
- important changes were not incorporated


boegel:
up to @epauwels... I'm strongly on favour of issuing a follow-up PR to actually tackle the typos/remarks rather than go back in time.

itkovian:
Until these are fixed, this leaves master in a non-pristine state. You could argue that the fixes were needed to bring it back to such a state, so if we merge this it will end up in a non-pristine state as well, but at least then people will not have to rebase onto something we know contains errors and typos that should have been fixed.

PR #119, opened 2016-04-15
Intro to linux

ehiggs:
@boegel and I gave a class for new Linux users. I converted the MediaWiki document we wrote to LaTeX and added it here. I tried to make it more general for other VSC sites so other groups can use it too if they don't already have such a document.

### DIFF
--- a/intro-Linux/ch_getting_started.tex
+++ b/intro-Linux/ch_getting_started.tex
@@ -3,7 +3,7 @@
 \section{Logging in}
 
 To get started with the HPC-UGent infrastructure, you need to obtain a VSC
-account, see http://hpc.ugent.be/userwiki/index.php/User:VscRequests .
+account, see \url{http://hpc.ugent.be/userwiki/index.php/User:VscRequests}.
 
 \strong{Keep in mind that you must keep your private key to yourself!}
 
@@ -13,7 +13,7 @@ key to open the door and get access to the HPC infrastructure. \strong{Anyone
 who has your key can use your VSC account!}
 
 Details on connecting to the HPC infrastructure are available at
-\usrl{http://hpc.ugent.be/userwiki/index.php/User:VscConnect}.
+\url{http://hpc.ugent.be/userwiki/index.php/User:VscConnect}.
 
 \section{Getting help}
 
@@ -30,7 +30,7 @@ To get help:
 \end{prompt}
 
   \item use Google
-  \item contact hpc@ugent.be in case of problems or questions (even for basic
+  \item contact \hpcinfo in case of problems or questions (even for basic
     things!)
 \end{enumerate}
 
@@ -40,7 +40,7 @@ Sometimes when executing a command, an error occurs. Most likely there will be
 error output or a message explaining you this. Read this carefully and try to
 act on it. Try googling the error first to find any possible solution, but if
 you can't come up with something in 15 minutes, don't hesitate to mail
-\email{hpc@ugent.be}.
+\hpcinfo.
 
 
 \section{Basic terminal usage}
--- a/intro-Linux/ch_getting_started.tex
+++ b/intro-Linux/ch_getting_started.tex
@@ -1,0 +1,185 @@
+\chapter{Getting Started}
+
+\section{Logging in}
+
+To get started with the HPC-UGent infrastructure, you need to obtain a VSC
+account, see http://hpc.ugent.be/userwiki/index.php/User:VscRequests .
+
+\strong{Keep in mind that you must keep your private key to yourself!}
+
+You can look at your public/private key pair as a lock and a key: you give us
+the lock (your public key), we put it on the door, and then you can use your
+key to open the door and get access to the HPC infrastructure. \strong{Anyone
+who has your key can use your VSC account!}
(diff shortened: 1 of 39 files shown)


intro-Linux/ch_getting_started.tex
```diff
@@ -0,0 +1,189 @@
...
+
+\strong{Keep in mind that you must keep your private key to yourself!}
+
+You can look at your public/private key pair as a lock and a key: you give us
+the lock (your public key), we put it on the door, and then you can use your
+key to open the door and get access to the HPC infrastructure. \strong{Anyone
+who has your key can use your VSC account!}
+
+Details on connecting to the HPC infrastructure are available at
+\usrl{http://hpc.ugent.be/userwiki/index.php/User:VscConnect}.
```
wpoely86: line is too long (needs wrapping/breaking)

intro-Linux/ch_getting_started.tex
```diff
@@ -0,0 +1,189 @@
...
+    ``info'' and ``man'' commands (use ``q'' to exit).
+
+\begin{prompt}
+  help cd
+  info ls
+  man cp
+\end{prompt}
+
+  \item use Google
+  \item contact hpc@ugent.be in case of problems or questions (even for basic
```
wpoely86: make all those `hpc@ugent.be` an actual link?

intro-Linux/ch_getting_started.tex
```diff
@@ -0,0 +1,189 @@
...
+They can be thought of as placeholders for things we need to remember.
+
+For example, to print the path to your home directory, we can use the shell variable named ``HOME'':
+
+\begin{prompt}
+%\shellcmd{echo \$HOME}%
+/user/home/gent/vsc400/vsc40000
+\end{prompt}
+
+This prints the value of this variable.
```
wpoely86: maybe says something that a variable is always prefixed by `$`?

intro-Linux/title.tex
```diff
@@ -0,0 +1,130 @@
...
+\begin{tabular}{ >{\centering\arraybackslash}m{0.25\textwidth}  >{\centering\arraybackslash}m{0.05\textwidth}  >{\centering\arraybackslash}m{0.20\textwidth}  >{\centering\arraybackslash}m{0.2\textwidth}} \\
+\includegraphics*[width=0.2\textwidth, height=0.7in, keepaspectratio=true]{logo_auha} & \multicolumn{2}{ >{\centering\arraybackslash}m{.2\textwidth} }{\includegraphics*[width=0.2\textwidth, height=0.7in,, keepaspectratio=true]{logo_akuleuven}} & \includegraphics*[width=0.2\textwidth, height=0.7in,, keepaspectratio=true]{logo_auhl} \\
+\multicolumn{2}{ >{\centering\arraybackslash}m{.32\textwidth} }{\includegraphics*[width=0.3\textwidth, height=0.7in, keepaspectratio=true]{logo_augent}} & \multicolumn{2}{ >{\centering\arraybackslash}m{.38\textwidth} }{\includegraphics*[width=0.3\textwidth, height=0.7in, keepaspectratio=false]{logo_uab}} \\
+\end{tabular}
+\end{center}
+
+
+\strong{Audience:}
+This document is a hands-on guide for using the Linux command line in the
+context of the \string{\university} \hpc infrastructure. The
```
wpoely86: this renders to: ` of the {UGent HPC infrastructure. `

intro-Linux/ch_getting_started.tex
ehiggs: See line 106.

klust:
Ewan,

I've merged your contribution so that I could proceed merging other pull requests also and doing further updates without having to rebase all the time, but I have two questions remaining:

1) Do you intend to make this a client-specific document since you have put it in all_coc_os rather than all_doc_noos in the Makefile? For now, the only dependency on the client system is in the title page, the remainder of the text is generic for all clients. Since I don't think we want to discuss clients to connect over ssh - this is already done in the intro-HPC document - I guess this document will not depend on the client OS but only on the site (for references to helpdesks etc.). So I think title.tex and the Makefile can be changed again. I can make these modifications if you have no time.

2) I see that you still talk about the -q-option for PBS to specify the queue. Is this something that is still used at Ghetn? Both in Leuven and Antwerp we currently discourage our users to specify the queue themselves as the Torque/MOAB combo is perfectly capable to route the job to the correct queue based on the other specifications (and both in Leuven and Antwerp we actually use partitions if that would not be the case). If you also don't use this at Ghent anymore, it might be better to remove it from the Linux tutorial.

Regards,

Kurt

boegel:
@klust Ewan (@ehiggs) no longer works in the HPC-UGent team since May 1st, so let me answer your questions.

1) yes, I think the intention is to make this a generic document, since the focus is on the usage of Linux, not on connecting to the HPC infrastructure; this is indeed already covered elsewhere

2) I agree that the use of `-q` should be discouraged (as agreed in the Common User Experience project a while ago), so yes, we should remove the use of `-q` from the document; we try to avoid telling our users to submit to a particular queue, this slipped through

PR #130, opened 2017-02-03
Various fixes

JensTimmerman:


intro-HPC/ch_account.tex
```diff
@@ -282,9 +285,9 @@ \section{Applying for the account}
 \includegraphics*[width=5.73in, height=4.00in, keepaspectratio=false]{ch2-browser-authenticate}
 \end{center}
 
-Select ``\university'' in the dropdown box.
+Select ``\university'' in the dropdown box and optionaly select Save my preference and permanently.
```
boegel: use quotes around `Save my preference` and `permanently`?

JensTimmerman: That makes it awkward with the `'permanently.` I think, but ok.

boegel: it reads weird without the quotes, imho

PR #134, opened 2017-04-24
link to request additional quota on account page

JensTimmerman:
Created from correct fork.


intro-HPC/ch_running_jobs_with_input_output_data.tex
```diff
@@ -231,7 +231,11 @@ \subsection{Your data directory (\$VSC\_DATA)\label{subsec:data-directory}}
 \ifgent
 If you are running out of quota on your \$VSC\_DATA filesystem you can
 request a VO (Virtual organisation). VO members get an extra DATA filesystem: \$VSC\_DATA\_VO)
-We can give higher quota values for these filesystems. Don't hesitate to contact the \hpcTeam staff.
+We can give higher quota values for these filesystems. VO administrators can request additional quota
+at the "Edit VO" page \url{https://account.vscentrum.be/django/vo/edit}.
+Do note that additional quota is an upper bound that can be raised, but this is not
+a guarantee you will be able to use the additional quota since we do oversubscribe our storage
+to achieve optimal usage.
```
boegel: Hmm, I find this last bit is confusing... Not sure if this is worth mentioning, since I doubt we'd let the filesystem run 100% full?

JensTimmerman: It could happen, and we want to make this clear, better inform users about this so they are not under the false impression that everyone can use all the quota they receive.

boegel: The existing paragraph creates more confusion rather than making things clear imho...

wdpypere:
@boegel make a suggestion here.

intro-HPC/ch_running_jobs_with_input_output_data.tex
```diff
@@ -231,7 +231,11 @@ \subsection{Your data directory (\$VSC\_DATA)\label{subsec:data-directory}}
 \ifgent
 If you are running out of quota on your \$VSC\_DATA filesystem you can
 request a VO (Virtual organisation). VO members get an extra DATA filesystem: \$VSC\_DATA\_VO)
```
boegel: `An extra directory on the DATA filesystem is created for a VO: \$VSC\_DATA\_VO`

intro-HPC/ch_running_jobs_with_input_output_data.tex
boegel: I would just drop this last part, to be honest... In practice, we're probably going to start taking measures when available storage space becomes so limited that users are being affected, either by asking users with large volumes of data to clean up or move some of their data, or by expanding the storage.

There's a theoretical possibility that the DATA filesystem gets filled to the brim and users won't be able to use all of their quota, but I strongly doubt that will ever happen in practice, since we're keeping a close eye on it...

JensTimmerman: Whut? Why not be 100% clear and accurate? 
let me remind you what happened last time I heard someone saying: practice learns this will never happen...

intro-HPC/ch_running_jobs_with_input_output_data.tex
JensTimmerman: we should be consistent about how we name things yes. In an upcoming data-policy we are currently opting for 'data partition' or 'partition' instead of filesystem/fileset/storage system. But that's a different issue to be handled separately from this PR

intro-HPC/ch_running_jobs_with_input_output_data.tex
boegel: If you want 100% accurate, you should also discuss hardware problems, possibly combined with software bugs (this may ring a bell or two), that may potentially destroy their data, or cosmic rays that may cause bit flips (it happens, see https://blogs.oracle.com/ksplice/attack-of-the-cosmic-rays).

Documentation should be clear & concise, not try to cover all potential "but-if's", especially not ones we should try hard to avoid from ever occurring (especially since operating at 90% would already be considered an important problem)...

Imho, this paragraph will only raise questions & concerns.

intro-HPC/ch_running_jobs_with_input_output_data.tex
boegel: I wasn't really pointing out an inconsistency, the `extra DATA filesystem` was simply incorrect, all they get is a directory on an existing filesystem, we don't create a new filesystem for them.

Technically, I'm not sure 'partition' is 100% correct. Filesystem is though, and it's a fairly common term, just like 'directory', while partition is less common and may raise questions (but maybe I'm biased, I haven't used Windows for a long time).
(What are the `D:`, `E:` additional locations next to `C:` called in Windows?)

That they're actually getting a fileset on a filesystem is irrelevant to the users, all they see and need to care about is a directory...

intro-HPC/ch_running_jobs_with_input_output_data.tex
itkovian: @boegel That is not merely a technical possibility. It has occurred. Multiple times. Whenever we hit 85%, or higher, we've asked people to clean up their stuff. Most of those people did not have a (>85%) full VO, quota-wise. So they were not even able to use that, and since they were big consumers of storage space, we asked them first to clean up ...

boegel: Suggestion for a better paragraph (it should be a *new* paragraph imho, not a few sentences at the end of the existing one):

```
Note that we may oversubscribe each shared filesystem, meaning that the total amount of quota across
all VOs may exceed the total size of the filesystem. Because of this, you should consider the available quota
as an upper limit of the amount of storage space that is available to you; you may not be able to use it all
if the filesystem does not provide a sufficient amount of free space.
```

intro-HPC/ch_running_jobs_with_input_output_data.tex
boegel: @JensTimmerman This line should change too, `an extra DATA filesystem` is just wrong imho...

Also, please sync your branch with current `master` and run the `check-version.sh` to bump the version.

JensTimmerman:
@boegel any idea how I fix this 'INTRO_HPC_EXPECTED' having the wrong value? Can this test be removed completely?

intro-HPC/ch_running_jobs_with_input_output_data.tex
JensTimmerman: again, see #158

JensTimmerman: @boegel why don't the tests do this? They merge with master, no?

JensTimmerman:
@boegel keeps failing
```
 ./check-version.sh                    
ERROR: Usage: ./check-version.sh <directory>
 ./check-version.sh intro-HPC/ 
Current version for intro-HPC/: 20171020.01 (found in intro-HPC//title.tex)
ERROR: Failed to determine current checksum for 'intro-HPC/'
```

is this documented somewhere?

boegel:
@JensTimmerman Please sync with current `master` and re-run `./check-version.sh intro-HPC`

JensTimmerman:
@boegel  I just did that

JensTimmerman:
@boegel travis log ends after 50 mins with 
``` 
Installing [2042/2398, time/total: 48:13/56:09]: tagpair [69k]
Installing [2047/2398, time/total: 48:18/56:10]: tcldoc [500k]
Installing [2048/2398, time/total: 48:19/56:10]: tcolorbox [4599k]
```

boegel:
@JensTimmerman Test restarted, may have been fallout from Travis outage (see https://www.traviscistatus.com/incidents/c1s4dlyxd4lf)

PR #132, opened 2017-03-06
added md5 and sha256 fingerprints of all 3 keytypes

JensTimmerman:

PR #138, opened 2017-07-07
fix warnings/errors reported for texlive.profile

boegel:
Current build of `master` is broken, since #132 was merged, this should fix it.

```
$ ./install-tl -profile $TRAVIS_BUILD_DIR/texlive.profile
...
Profile key `in_place' is now ignored, please remove it.
Profile key `option_menu_integration' is now ignored, please remove it.
The profile references a non-existing collection: collection-langindic
Exiting.
...
The command "./install-tl -profile $TRAVIS_BUILD_DIR/texlive.profile" failed and exited with 1 during .
```

I only removed those lines that referenced to "non-existing collections" + ignored ones.

### DIFF
--- a/texlive.profile
+++ b/texlive.profile
@@ -19,11 +19,7 @@ collection-fontsrecommended 1
 collection-fontutils 1
 collection-formatsextra 0
 collection-games 0
-collection-genericextra 1
-collection-genericrecommended 1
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

PR #141, opened 2017-08-21
rename 'Good Practices' chapter to 'Best Practices'

boegel:
I think the term "Best Practices" is more commonly used...

The intention of the newly added sections for software-specific best practices is to document best practices specific to the use on HPC(/VSC) systems, *not* to replicate/replace existing user documentation...

I intend to populate the placeholder sections for EasyBuild & OpenFOAM very soon in upcoming PRs.

For EasyBuild, I will port the documentation we currently have on our userwiki, which augments the EasyBuild documentation at http://easybuild.readthedocs.io/ with (limited) VSC/site-specific information.

For OpenFOAM, I will work out best practices together with experienced HPC-UGent OpenFOAM users, see also https://www.ugent.be/hpc/en/support/openfoamusermeeting.

Other tools that could/should be documented here as well in the (near?) future:

* Lmod (ported from https://www.vscentrum.be/cluster-doc/software/modules-lmod)
* MATLAB (ported from from HPC-UGent userwiki)
* Worker (ported from https://www.vscentrum.be/cluster-doc/running-jobs/worker-framework)
* Singularity (ported from HPC-UGent userwiki)


intro-HPC/HPC.tex
```diff
@@ -1,88 +0,0 @@
-\documentclass[11pt,a4paper,oneside]{book}
```
wpoely86: this file is referenced in the style guide, please change it there too.

PR #144, opened 2017-08-23
use datestamp as version + check for version bump when intro-HPC/*.tex files are changed

boegel:
The version of `intro-HPC` has been `1.1` for quite a while now...

I propose to start using a datestamp (+ minor 2-digit version number) as a version, which is more meaningful; for example: `20170823.01`, i.e. the first update done on Aug 23rd 2017.

In addition, we should have automated checking of whether the version number was indeed bumped when changes were made, to avoid overlooking this in the future.

The `check-version.sh` script included here has two purposes:

* it can be used to automatically bump the version if any changes were made to the `*.tex` files
* it can be used in the Travis config file to verify whether the version was indeed bumped when changes were made to `*.tex`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ env:
       # and replace it below
       - secure: 2p42CdapsQR6xkcR/cnqngbwwYqol7QKosxa6n7r4HhKFw7pcaSRWa/5fLOKIunMxW56UfQgTHCecGAU5/oWS8jmEVZYpBThynQBsHBt4xtPw4RUJzAwn6BmWeKX4RHKE6CCcscxAa1Q9vxBWJ0/yRLSP/ikfj6cAH33ga00dFA=
 before_install:
+    # verify SHA256 checksums to ensure version is bumped when changes are made
+    # if this check fails, you should run ./check-version.sh intro-HPC and commit the changes made by the script
+    - INTRO_HPC_EXPECTED="20170823.01 965d34300067db2e0230c9f4281f3c408daa5c26b860c3e61669cf3069a2b73c"
+    - ./check-version.sh intro-HPC | grep "Checksum for intro-HPC is still valid"
     - cd $HOME
     - travis_retry wget http://mirror.ctan.org/systems/texlive/tlnet/install-tl-unx.tar.gz
     - tar -xzf install-tl-unx.tar.gz ; cd install-tl-*/
--- a/check-version.sh
+++ b/check-version.sh
@@ -51,8 +51,7 @@ if [ -z $checksum_tool ]; then
     exit 4
 fi
 
-mkdir -p /tmp/$USER
-tmpfile=$(mktemp /tmp/$USER/XXXXX)
+tmpfile=$(mktemp /tmp/${USER}_XXXXX)
 
 find $dir -name '*.tex' | sort -u | xargs $checksum_tool > $tmpfile
 new_checksum=$($checksum_tool $tmpfile | cut -f1 -d' ')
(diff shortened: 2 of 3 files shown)


check-version.sh
```diff
@@ -0,0 +1,96 @@
...
+if [ "x$checksum_tool" == "xshasum" ]; then
+    checksum_tool="shasum -a 256"
+fi
+
+if [ -z $checksum_tool ]; then
+    echo "ERROR: No checksum tool found!" >&2
+    exit 4
+fi
+
+mkdir -p /tmp/$USER
```
wpoely86: euh, why this? Isn't `mktemp` sufficient?

check-version.sh
```diff
@@ -0,0 +1,96 @@
...
+        new_minor_ver=$(expr $curr_minor_ver + 1)
+        new_version=${datestamp}.$(printf "%02d" ${new_minor_ver})
+    else
+        new_version="${datestamp}.01"
+    fi
+
+    echo "New version: $new_version"
+
+    # inject new version
+    sed -i.bak "s/Version ${curr_version}/Version ${new_version}/g" $curr_version_file
```
wpoely86: why `g`?

check-version.sh
```diff
@@ -0,0 +1,96 @@
...
+
+    # need to recompute checksum after injecting new version...
+    find $dir -name '*.tex' | sort -u | xargs $checksum_tool > $tmpfile
```
wpoely86: if you do this more than once, put it in  a function.

.travis.yml
```diff
@@ -8,6 +8,10 @@ env:
       # and replace it below
       - secure: 2p42CdapsQR6xkcR/cnqngbwwYqol7QKosxa6n7r4HhKFw7pcaSRWa/5fLOKIunMxW56UfQgTHCecGAU5/oWS8jmEVZYpBThynQBsHBt4xtPw4RUJzAwn6BmWeKX4RHKE6CCcscxAa1Q9vxBWJ0/yRLSP/ikfj6cAH33ga00dFA=
 before_install:
+    # verify SHA256 checksums to ensure version is bumped when changes are made
+    # if this check fails, you should run ./check-version.sh intro-HPC and commit the changes made by the script
+    - INTRO_HPC_EXPECTED="20170823.01 965d34300067db2e0230c9f4281f3c408daa5c26b860c3e61669cf3069a2b73c"
+    - ./check-version.sh intro-HPC | grep "Checksum for intro-HPC is still valid"
```
wpoely86: why the `grep`?

boegel: Because the script always exits with 0, even if it replaced the version/checksum, I don't consider that to be an error?

check-version.sh
boegel: yeah, not needed, indeed

check-version.sh
boegel: force of habit, will fix

check-version.sh
boegel: makes sense, thx

boegel:
@wpoely86 please rereview?

check-version.sh
```diff
@@ -51,11 +51,14 @@ if [ -z $checksum_tool ]; then
     exit 4
 fi
 
-mkdir -p /tmp/$USER
-tmpfile=$(mktemp /tmp/$USER/XXXXX)
-
-find $dir -name '*.tex' | sort -u | xargs $checksum_tool > $tmpfile
-new_checksum=$($checksum_tool $tmpfile | cut -f1 -d' ')
+function compute_checksum() {
+    tmpfile=$(mktemp /tmp/${USER}_XXXXX)
```
wpoely86: why hardcode the path? Let `mktemp` handle it completely? What if the user has a different `$TMPDIR` variable?

PR #145, opened 2017-08-24
Include ssh-add info

### LINKED ISSUE #142: include ssh-add info into intro-HPC-linux doc
The most recent GNU/Linux distros use a SSH agent or compatible service like `gnome-keyring-daemon`  to manage the user SSH keys. ` ssh-keygen`   command is not enough to connect to the cluster, the key should be loaded after that with `ssh-add`  command. (This was tested in Ubuntu 16.04)

After that the user will be able to connect to the login nodes as usual with his/her ssh key.
We should add a `ssh-add`   example into chapter 2 (section 2.1.1) to explain how to use the ssh key agent from a linux distro.

alvarosimon:
* Fix #142

### DIFF
--- a/intro-HPC/ch_account.tex
+++ b/intro-HPC/ch_account.tex
@@ -263,25 +263,47 @@ Your public key has been saved in /home/user/.ssh/id_rsa.pub.
 
 \iflinux
 
-\subsection{Using a SSH agent (optional)}
+\subsection{Using an SSH agent (optional)}
 \label{sec:using-ssh-agent-with-openssh}
-  The most recent GNU/Linux distributions include by default a SSH agent 
-  (``gnome-keyring-daemon'' in most cases) to keep and manage our SSH keys. 
-  For those distributions the user \strong{must} include the new key/s into 
-  the SSH manager keyring to be able to connect to the HPC cluster.
-  This could be done thanks to ``ssh-add'' command.
+  The most recent GNU/Linux distributions include by default an SSH agent 
+  (``gnome-keyring-daemon'' in most cases) to keep and manage the user SSH keys.
+  If you use one of these distributions you \strong{must} include the new key/s into 
+  the SSH manager keyring to be able to connect to the HPC cluster. If not,
+  SSH client will display an error message (see \autoref{ch:preparing-the-environment}) 
+  similar to this:
+  
+  \begin{flattext}
+   Agent admitted failure to sign using the key.
+   Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
+  \end{flattext}
 
+  This could be fixed using the ``ssh-add'' command.
+  You can include the new private key/s identities in your keyring with:
 \begin{prompt}
-Include the new key/s in our keyring:
 %\shellcmd{ssh-add}%
-Check that our key is available from the keyring:
+\end{prompt}
+
+\begin{tip}
+  Without extra options ``ssh-add'' adds any key located at ``~/.ssh''
+  directory, but you can specify the private key location path as argument,
+  as example ``ssh-add /path/to/my/id\_rsa''.
+\end{tip}
+
+Check that your key is available from the keyring with:
+\begin{prompt}
 %\shellcmd{ssh-add -l}%
 \end{prompt}
   
-  After these changes the key agent will keep our SSH key to connect 
-  to the clusters.
+  After these changes the key agent will keep your SSH key to connect 
+  to the clusters as usual. 
+\begin{tip}
+  You should execute ``ssh-add'' command again if you generate a new SSH key.
+\end{tip}
 
-\fi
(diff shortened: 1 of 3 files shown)


intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,28 @@ \section{Getting ready to request an account}
   \strong{Without your key pair, you won't be able to apply for a personal VSC account.}
 \fi %macORlinux
 
+\iflinux
+
+\subsection{Using a SSH agent (optional)}
```
boegel: *an* SSH agent

intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,28 @@ \section{Getting ready to request an account}
...
+\label{sec:using-ssh-agent-with-openssh}
+  The most recent GNU/Linux distributions include by default a SSH agent 
+  (``gnome-keyring-daemon'' in most cases) to keep and manage our SSH keys. 
+  For those distributions the user \strong{must} include the new key/s into 
+  the SSH manager keyring to be able to connect to the HPC cluster.
+  This could be done thanks to ``ssh-add'' command.
+
+\begin{prompt}
+Include the new key/s in our keyring:
+%\shellcmd{ssh-add}%
```
wpoely86: and the key as an argument?

intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,28 @@ \section{Getting ready to request an account}
...
 
+\iflinux
+
+\subsection{Using a SSH agent (optional)}
+\label{sec:using-ssh-agent-with-openssh}
+  The most recent GNU/Linux distributions include by default a SSH agent 
+  (``gnome-keyring-daemon'' in most cases) to keep and manage our SSH keys. 
+  For those distributions the user \strong{must} include the new key/s into 
+  the SSH manager keyring to be able to connect to the HPC cluster.
+  This could be done thanks to ``ssh-add'' command.
```
boegel: thanks to -> using the

also mention the error you get on recent distros if you don't use `ssh-add`, so people can search for it easily?

intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,28 @@ \section{Getting ready to request an account}
...
+
+\begin{prompt}
+Include the new key/s in our keyring:
+%\shellcmd{ssh-add}%
+Check that our key is available from the keyring:
+%\shellcmd{ssh-add -l}%
+\end{prompt}
+  
+  After these changes the key agent will keep our SSH key to connect
```
boegel: our -> your

also mention that this needs to be done again after logging out/rebooting?

intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,28 @@ \section{Getting ready to request an account}
...
+\begin{prompt}
+Include the new key/s in our keyring:
+%\shellcmd{ssh-add}%
+Check that our key is available from the keyring:
+%\shellcmd{ssh-add -l}%
+\end{prompt}
+  
+  After these changes the key agent will keep our SSH key to connect 
+  to the clusters.
+
```
wpoely86: I would add a link to some documentation how to work with an ssh agents? Like https://wiki.archlinux.org/index.php/SSH_keys#SSH_agents

alvarosimon:
@boegel @wpoely86 I have included most of your remarks, please review the pr again. I want to keep this info as simple as possible just to avoid any confusion to the new users.

intro-HPC/ch_account.tex
boegel: @alvarosimon It's probably good to mention that `ssh-add` only works if the private key is in the default location at `~/.ssh/id_rsa`

if not, you need to do `ssh-add <path/to/private/key>`

intro-HPC/ch_account.tex
```diff
@@ -259,6 +259,50 @@ \section{Getting ready to request an account}
...
+  \end{flattext}
+
+  This could be fixed using the ``ssh-add'' command.
+  You can include the new private key/s identities in your keyring with:
+\begin{prompt}
+%\shellcmd{ssh-add}%
+\end{prompt}
+
+\begin{tip}
+  Without extra options ``ssh-add'' adds any key located at ``~/.ssh''
```
wpoely86: did you test this when the user has more then 3 keys? ssh will only try 3 keys and then give up. So it might not be a good idea to add all keys to the agent.

alvarosimon: We should keep this as simple as possible imho. This document just describes how to generate a ssh key with `ssh-keygen`   without extra options in the previous section, that will create a `id_rsa`  key by default. We can include a lot of examples about `ssh-keygen`  and `ssh-add`   but I think that this is out of the scope of this document, we provide external links if the user wants to try something else. I think that this is enough for a regular user 😄

PR #149, opened 2017-09-11
flesh out best practices for OpenFOAM

boegel:
preview available at http://users.ugent.be/~kehoste/intro-HPC-mac-gent.pdf, see section 13.5 (pages 116-124)

I've reorganised the sections a bit so we can have numbered subsections per application/tool.

Two subsection still need to be fleshed out, I will do that after the OpenFOAM user meeting this week (https://www.ugent.be/hpc/en/support/openfoamusermeeting).


klust:
I actually wonder where that (very useful) information should go. Should it be in the tutorial text, a separate "Best practices" PDF or just on the web site? The tutorial is becoming so thick that at some point it will scare away users, also because it will start to contain so much information that is not relevant for him/her that the relevant information will be overlooked.

I personally believe more in further enhancing the "Using software" section of the User Portal on our website for this kind of information. 

And aren't best practices for EasyBuild not something that is more for system administrators and not for regular users?

That being said, I am very much in favour of adding documentation on how to use mympirun earlier in the tutorial, e.g., in section 7.4 on MPI, in particular since it is a very useful tool to start hybrid MPI/OpenMP programs (or other hybrids).

JensTimmerman:
@boegel please fix conflichts
@klust I agree, we could perhaps split the pdf up into 2 pieces, an intro, and an advanced part?

klust:
I suggest to have a brief discussion during a coffee break during the hands-on in a few weeks, unless this would even be a separate topic. The tutorial is also getting a bit "long in the tooth" in certain sections, and keeping it up to date seems to be even harder in practice than keeping a web site up to date, and users are even less likely to take that document again as they are searching on a web site so we should from time to time seriously reconsider which information should go where.

boegel:
To revive this a bit: the current proposal is to add a "best practices" part III with software-specific documentation, see #196, where a chapter dedicated to OpenFOAM can be included.

boegel:
@redfast00 Can you look into moving this into the new Part III now that #265 is merged?

Also under `\ifgent` please, and go through the documentation and just kick out any "TODO" sections, we can add those back later...

redfast00:
@boegel this has been done #304, can this be closed?

PR #164, opened 2017-12-15
update contact information for Brussels + various small fixes/improvements

smoors:


glossary.tex
```diff
@@ -46,7 +46,7 @@
 \newglossaryentry{L1d}
 {
   name={L1d},
-  description={Level 1 data cache, often called \textbf{primary cache}, is a static memory integrated with processor core that is used to store data recently accessed by a processor and also data which may be required by the next operations.},
+  description={Level 1 data cache, of$ten called \textbf{primary cache}, is a static memory integrated with the CPU core that is used to store data recently accessed by a processor and also data which may be required by the next operations},
```
boegel: What's with the `$` in `of$ten`?

`Processor` should also be mentioned in the glossary, as a synonym for 'CPU'?

We usually talk about a "multi-core processor" not a "multi-core CPU" (since the 'unit' part of CPU is confusing w.r.t. a core which is a single compute 'unit').

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -18,15 +18,19 @@ \chapter{Running batch jobs}
 nodes} of the cluster. There you can prepare the work you want to get done on
 the cluster by, e.g., installing or compiling programs, setting up data sets,
 etc. The computations however, should not be performed on this login node. The
-actual work is done on the cluster's \strong{compute nodes}. These compute
-nodes are managed by the job scheduling software (Moab) and a Resource Manager
-(TORQUE), which decides when and on which compute nodes the jobs can run. It is
-usually not necessary 
+actual work is done on the cluster's \strong{compute nodes}. Each compute node
+contains a number of CPU \strong{cores}. The compute nodes are managed by the
```
boegel: "processor cores"?

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -18,15 +18,19 @@ \chapter{Running batch jobs}
...
+job scheduling software (Moab) and a Resource Manager (TORQUE), which decides
```
boegel: Should be `Torque` and `MOAB`, I think...

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -18,15 +18,19 @@ \chapter{Running batch jobs}
...
+when and on which compute nodes the jobs can run. It is usually not necessary
 %% TODO: where?
 %(and in some cases not permitted)
 to log on to the
-compute nodes directly. Users can (and should) monitor their jobs
-periodically as they run, but do not have to remain logged in the entire
+compute nodes directly
+\ifbrussel
+and is only allowed on the node on which your job is running
```
boegel: "on the nodes where you have a job running"

This is also true for (at least) Gent.

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -428,29 +432,32 @@ \subsection{Generic resource requirements}
 \end{prompt}
 
 For the simplest cases, only the amount of maximum estimated execution time
-(called ``walltime'') is really important. Here, the job will not require more
-than 2 hours, 30 minutes to complete. As soon as the job would take more time,
+(called ``walltime'') is really important. Here, the job requests a reservation of
```
boegel: not sure if the term 'reservation' should be used here, since that has a very specific meaning?

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -428,29 +432,32 @@ \subsection{Generic resource requirements}
...
+2 hours, 30 minutes. As soon as the job starts to take more time,
```
boegel: "exceeds the requested (wall?) time"?

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -428,29 +432,32 @@ \subsection{Generic resource requirements}
...
 it will be ``killed'' (terminated) by the job scheduler.  There is absolutly no
 harm if you \emph{slightly} overestimate the maximum execution time. \\
 
 \begin{prompt}
 %\shellcmd{qsub -l mem=4gb}%
 \end{prompt}
 
-The job requires no more than 4 GB of memory. \\
+The job requests a reservation of 4 GB of RAM memory. As soon as the job tries to use more memory,
+it will be ``killed'' (terminated) by the job scheduler.  There is absolutly no
```
boegel: `absolutely`

Maybe this is overstated a bit, since it may have a negative effect on queuing time (requested more resources may mean that the job has the wait longer).

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -428,29 +432,32 @@ \subsection{Generic resource requirements}
...
+harm if you \emph{slightly} overestimate the requested memory. \\
+ \\
 
 \begin{prompt}
 %\shellcmd{qsub -l nodes=5:ppn=2}%
 \end{prompt}
 
-The job requires 5 compute nodes with two cores on each node (ppn stands for
-``processors per node'', where processor is used to refer to individual cores). \\
+The job requests a reservation of 5 compute nodes with two cores on each node (ppn stands for
```
boegel: Same here, I would avoid the term 'reservation'?

intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -428,29 +432,32 @@ \subsection{Generic resource requirements}
...
+``processors per node'', where "processors" here actually means "CPU cores"). \\
```
boegel: "processor cores"?

glossary.tex
smoors: Thanks for the remarks.
I prefer CPU, because processor is more general, it can mean GPU as well as CPU. It's clear that we mean CPU here.
I ack the possible confusion, that is why I explained in the glossary that a CPU can contain more than one core.
If you agree, I will make the distinction between processor and CPU in the glossary.

intro-HPC/ch_running_batch_jobs.tex
smoors: see above.

intro-HPC/ch_running_batch_jobs.tex
smoors: Yes, this was a pre-existing error.

intro-HPC/ch_running_batch_jobs.tex
smoors: ok, will add Gent

intro-HPC/ch_running_batch_jobs.tex
smoors: You have a point here, I will remove it.

intro-HPC/ch_running_batch_jobs.tex
smoors: That's more clear, will change.

intro-HPC/ch_running_batch_jobs.tex
smoors: ok.

intro-HPC/ch_running_batch_jobs.tex
smoors: see my reply above.

glossary.tex
boegel: I never looked at it that way, but I guess you're right, strictly speaking. When we say 'processor' it's usually a synonym for CPU.

I can see why you would like to make the distinction and would use 'processor' as an umbrella  term for CPU/GPU though, so I'm fine with that. As long as we use it consistently.

intro-HPC/ch_running_batch_jobs.tex
smoors: Hm, it seems it is correct like it is.

glossary.tex
smoors: OK, I will make sure it is consistent throughout the tutorial.

intro-HPC/ch_running_batch_jobs.tex
boegel: Hmm, OK, nevermind me then. :)

glossary.tex
```diff
@@ -33,6 +33,11 @@
   name={FTP},
   description={File Transfer Protocol, used to copy files between distinct machines (over a network.) FTP is unencrypted, and as such blocked on certain systems. SFTP or SCP are secure alternatives to FTP},
 }
+\newglossaryentry{GPGPU}
+{
+  name={GPGPU},
+  description={A general purpose graphical processing unit. A GPGPU is a consumable resource. A GPGPU is a GPU that performs non-specialized calculations that would typically be conducted by the CPU. A compute node may contain zero, one or more GPGPUs},
```
boegel: The `non-specialized` is a bit unclear to me here...

smoors: How about:  "A GPGPU is a GPU that is used for highly parallel general purpose calculations." ?

boegel: :+1:

PR #166, opened 2018-01-16
Update puttykeygen screenshots

alvarosimon:
Some users are a bit confusing regarding the current puttygen documentation. The new putty version shows different options.
We should update the putty keygen  screenshots to avoid this confusion.


boegel:
@alvarosimon Update looks excellent, but you'll need to update the version too using `./check-version.sh intro-HPC`.

alvarosimon:
@boegel done!

PR #167, opened 2018-01-29
VPN is no longer required (except for UA login nodes)

JensTimmerman:


intro-HPC/ch_preparing_the_environment.tex
```diff
@@ -189,7 +189,7 @@ \section{First Time connection to the \hpc}
   \subsection{Connect}
   \label{sec:connect}
 
-  The \hpc is only accessible from within the \university network, but you can
+  At some institutes the \hpc is only accessible from within the \university network, but you can
```
wpoely86: Is this still true? I know for us  (VUB) it is not the case, Leuven neither. Gent the same? I think only Antwerp might be? @klust ?

JensTimmerman: Hence the change in wording here.
I wanted to add an `\ifantwerp`, but it seems to break inside an 
`\ifmacorlinux`

JensTimmerman:
@wpoely86 any way of fixing `The job exceeded the maximum time limit for jobs, and has been terminated.`
Faster/smaller latex install?

wpoely86:
Yes, smaller install or try the cache in Travis.

What we can do is do only a very basic install of LaTeX and install just the package we use on top of that.

boegel:
@JensTimmerman Please sync with `master`, re-bump version so we can try again?

PR #174, opened 2018-03-27
layout-only commits

ngaro:
I have no affiliation with VSC, but I have a research-group filled with VSC users that somehow always manage to bother me with their problems.

I'm planning to use your docs to base my own docs upon, so here is something in return:
- _Directory not following standard directory layout_ : pretty much useless commit, someone just ignored the layout mentioned in the beginning for dirs
- _tree in utf-8_ : There is no place on this planet where you will still see 'tree' giving ASCII-output
- _better tab-key_ : The tab-symbol (on windows keyboards) instead of the word TAB on the key, I don't know how it looks on a Mac, please buy me one...
- Fix larger/smaller than in code and shell : should follow the default layout


boegel:
@ngaro Please run `./check-version.sh intro-HPC` to update the version & make Travis happy.

ngaro:
Tomorrow, AFK for now...

ngaro:
version bump

style.tex
```diff
@@ -20,7 +20,7 @@
 \usepackage{courier}
 \usepackage{xparse}
 \usepackage{longtable}
-\usepackage{menukeys}
+\usepackage[os=win]{menukeys}
```
JensTimmerman: would it be possible to make this option depend on the current version of the pdf you're building?

JensTimmerman:
looks good to me, @ngaro thank you very much!

ngaro:
Nice one, missed that macro.
A new pull-request arrives after lunch.

Some extra unrelated info:
- I'm changing my fork from you to a fork from Antwerpen.
I often see them face to face, they know you (I don't) and if i'm correct Stefan and Kurt have direct access to your repo.
Working like this will both speed up communication and improve results.

- You'll probably want to know that locally I only test my builds with "OS=linux SITE=antwerpen", all other builds are problems for Travis and reading the resulting pdfs for other builds will become a problem for antwerpen.

- You can pull and cherry pick from my master as much as you want, but I also have a couple of branches for myself with names starting with "cmt".
You'll sometimes see them pop up on github. I suggest staying away from them. Although my changes here shouldn't affect others builds, they are only useful for my users so it will just slow down your latex-compiling

boegel:
We have applied the changes that were still relevant as a part of #286, thanks for your contribution @ngaro, and sorry for letting it sit for so long...

PR #181, opened 2018-08-27
(Linux intro, getting started) Corrected typo, added explanation about env command #188

redfast00:


intro-Linux/ch_getting_started.tex
```diff
@@ -73,8 +73,7 @@ \subsection{Command history}
 
 \subsection{Stopping commands}
 If for any reason you want to stop a command from executing, press Ctrl-C. For
-example, if a command is taking too long and you want to rerun it with a
-different command.
+example, if a command is taking too long and you want to rerun it with different arguments.
```
boegel: `and you want` -> `, or you want`

intro-Linux/ch_getting_started.tex
```diff
@@ -94,13 +93,23 @@ \section{Variables}
...
 
 There are several variables already defined for you when you start your
-session, such as ``\$HOME'' which contains the path to your home directory. 
+session, such as ``\$HOME'' which contains the path to your home directory.
+
+For a full overview of defined environment variables in your current session,
+you can use the ``env'' command:
+
+\begin{prompt}
+%\shellcmd{env}%
```
boegel: It's generally better to push this through `sort`, i.e. `env | sort` (unsorted output makes it hard to look for stuff), and often useful to `grep` for certain things like `env | sort | grep VSC`.

Maybe mention that too, i.e.

i) use `env | sort` here, and explain in one sentence why `sort` is used
ii) also mention (below) that you can `grep` for certain variable names or values, and give a short example (e.g. `env | sort | grep VSC` to find all VSC-specific environment variables

Try to keep it short though...

intro-Linux/ch_getting_started.tex
```diff
@@ -94,13 +93,23 @@ \section{Variables}
...
+...
+HOME=%\homedir%
```
boegel: The `%homedir%` is going to confuse people, since that's not what they will be seeing.

Use the actual value, but anonimise it, like `vsc40000`. We probably need a macro for an anonimised VSC login in, since it'll be different for each site (`vsc4xxxx` for Ghent, etc.)

intro-Linux/ch_getting_started.tex
```diff
@@ -158,6 +167,8 @@ \subsection{Using non-defined variables}
 
 The moral here is: \strong{be very careful to not use empty variables unintentionally}.
 
+\strong{Tip for job scripts: use ``set -e -u'' to avoid using empty variables accidentally.}
```
boegel: That's a good tip, but you'll need to explain it a bit more: what does `-e` do, what does `-u` do, and give an output example (like for `echo $DOESNOTEXIST`).

intro-Linux/ch_getting_started.tex
redfast00: `%\homedir%` is a macro, and gets expanded (? not sure if this is the right word) to `/user/home/gent/vsc400/vsc4000`.

boegel: I overlooked that, my bad... If it gets expanded as expected in the rendered document, then it's OK.

intro-Linux/ch_getting_started.tex
```diff
@@ -3,7 +3,7 @@ \chapter{Getting Started}
 \section{Logging in}
 
 To get started with the HPC-UGent infrastructure, you need to obtain a VSC
-account, see \url{http://hpc.ugent.be/userwiki/index.php/User:VscRequests}.
+account, see \url{http://hpcugent.github.io/vsc_user_docs/pdf/intro-HPC-linux-gent.pdf#chapter.2}.
```
boegel: Maybe make this a bit more "user friendly", like this?

```
\href{<url>}{Chapter 2 of the VSC user manual}
```

Do check whether that renders as expected though, i.e. whether it's clear that's a link to another document or not...

intro-Linux/ch_getting_started.tex
```diff
@@ -94,13 +93,31 @@ \section{Variables}
 \subsection{Defining variables}
 
 There are several variables already defined for you when you start your
-session, such as ``\$HOME'' which contains the path to your home directory. 
+session, such as ``\$HOME'' which contains the path to your home directory.
```
boegel: @redfast00 Use `\texttt` here for `$HOME`?

intro-Linux/ch_getting_started.tex
```diff
@@ -158,6 +175,14 @@ \subsection{Using non-defined variables}
 
 The moral here is: \strong{be very careful to not use empty variables unintentionally}.
 
+\strong{Tip for job scripts: use \texttt{set -e -u} to avoid using empty variables accidentally.}
+
+The \texttt{-e} option will immediately stop the script if any command fails and
```
boegel: Using `-e` will not stop the script, it will result in the script stopping if ... ;)

Same below

intro-Linux/ch_getting_started.tex
```diff
@@ -158,6 +175,14 @@ \subsection{Using non-defined variables}
...
+the \texttt{-u} option will stop the script if when using empty variables.
+(see \url{https://ss64.com/bash/set.html} for a more detailed explanation and more options)
+
+
+
```
boegel: No need for all this whitespace?

intro-Linux/ch_getting_started.tex
```diff
@@ -3,25 +3,24 @@ \chapter{Getting Started}
 \section{Logging in}
 
 To get started with the HPC-UGent infrastructure, you need to obtain a VSC
-account, see \url{http://hpc.ugent.be/userwiki/index.php/User:VscRequests}.
-
+account, see \href{http://hpcugent.github.io/vsc_user_docs/pdf/intro-HPC-linux-gent.pdf#chapter.2}{Chapter 2 of the HPC manual}.
 \strong{Keep in mind that you must keep your private key to yourself!}
 
-You can look at your public/private key pair as a lock and a key: you give us
+You \OS can look at your public/private key pair as a lock and a key: you give us
```
boegel: Why is `\OS` included here?! Search & replace gone bad?

intro-Linux/ch_getting_started.tex
```diff
@@ -154,10 +170,17 @@ \subsection{Using non-defined variables}
 %\homedir%
 \end{prompt}
 
-To understand what's going on here, see the section on ``cd'' below.
+To understand what's going on here, see the section on \verb|cd| below.
 
 The moral here is: \strong{be very careful to not use empty variables unintentionally}.
 
+\strong{Tip for job scripts: use \texttt{set -e -u} to avoid using empty variables accidentally.}
```
boegel: Either use `\verb` or `\texttt`, don't mix both (or explain why you're mixing them).

Also open a separate PR for the style guide on using `\verb` or `'\texttt`

intro-Linux/ch_getting_started.tex
```diff
@@ -128,10 +144,10 @@ \subsubsection{Changing your prompt using ``\$PS1''}
 ~/test $
 \end{prompt}
 
-Note that ``\tilde'' is short representation of your home directory.
+Note that \verb|~| is short representation of your home directory.
```
boegel: Does it still render correctly when you use `~` rather than `\tilde`?

redfast00: I'm using `\verb|~|`, because if just `~` is used, it sticks to the next word.

intro-Linux/ch_getting_started.tex
redfast00: Yup, I done messed up

intro-Linux/ch_getting_started.tex
redfast00: I'll put it in the styleguide later, but short explanation: \verb can't be used in arguments (so not in strong, section, ...). \verb is better though, because you don't have to worry about escaping double hyphens,...

intro-Linux/ch_getting_started.tex
boegel: OK, but why prefer `~` over `\tilde`? Just wondering... "less typing" is a good reason to me ;)

intro-Linux/ch_getting_started.tex
boegel: Ah, you already did (#217), thanks.

boegel:
@redfast00 Looks good to go, just missing version update to make Travis give a green light...

PR #182, opened 2018-08-27
Linux chapter navigating

redfast00:


boegel:
@redfast00 These changes are good to go, but I think the Travis config doesn't run the version check yet on the Linux introduction, we should change that.

Can you look into that first (please make it a separate PR), see #184

redfast00:
@boegel #184 was merged,  do I now do the final commit (run version script)?

boegel:
@redfast00 Yes, sounds good to me.

intro-Linux/ch_navigating.tex
```diff
@@ -115,7 +116,7 @@ \section{Absolute vs relative file paths}
 with ``/''), which is also called the \emph{root} of the filesystem.
 
 Example: absolute path to your home directory:
-``%\homedir%''.
+``\homedir''.
```
boegel: @redfast00 Ah, we should probably have used `\texttt{\homedir}` here?

Please fix that in some other PR where it makes sense (one that deal with typos/formatting in another chapter)

PR #197, opened 2018-08-27
Make navigating easier by coloring links blue

redfast00:


boegel:
@redfast00 Can you include a screenshot to show how it looks before & after?

redfast00:
screen See the blue text in the window on the left? That's how it looks after. Now imagine it's black, that's the before.

boegel:
@redfast00 Would underlining as well make sense, you think? Like people are used to on websites (maybe I'm still stuck in the 90s though).

redfast00:
This is "hard", are regular blue links fine?

boegel:
Underlined links are probably ugly too, so let's go forward with this as is...

PR #221, opened 2018-08-28
Linux chapter filesystems

redfast00:


intro-Linux/ch_filesystems.tex
```diff
@@ -3,36 +3,8 @@ \chapter{Filesystems}
...
-\subsubsection{Local}
-
-When running jobs, you also have access to the local storage of a node (which
-you cannot access when logging in, as it is node-dependent).
-
-You access it through ``\$VSC\_SCRATCH\_NODE''.
-
-Inside jobs, a unique directory located in ``\$VSC\_SCRATCH\_NODE'' is made
-available via ``\$TMPDIR''.
+See \href{http://hpcugent.github.io/vsc_user_docs/pdf/intro-HPC-linux-gent.pdf#subsection.6.2.1}{Chapter 6.2.1 of the HPC manual}
```
boegel: @redfast00 This link will easily go outdated if the section numbers change, so we should use a proper label here and mention something like `relevant section of the HPC manual` (or perhaps a bit more descriptive).

Also, rather than hard linking to `linux-gent`, we should take into account the `$OS` and `$SITE` variables somehow...

redfast00: I just wrote a macro to generate the url (taking into account the site and OS), but there are no labels in the exported intro-HPC pdf (and even if there were, jumping to them isn't supported by all readers). Do I link to the section and put the title in the description? I think that would be best of both worlds: users can always look it up (even if the order of chapters changes) and in a lot of the cases will be immediately directed to the correct chapter (but if it changes, they arrive at the wrong page). Another option is to just give the pdf link without setting a section/chapter. What do you want me to do?

redfast00: This PR is blocked by #223 for the links to the manuals.

redfast00:
@boegel

intro-Linux/ch_filesystems.tex
```diff
@@ -46,26 +18,16 @@ \subsubsection{VO storage}
...
+command can come in useful:
 
 \begin{prompt}
 %\shellcmd{du -sh test}%
 59M   test
 \end{prompt}
 
-Note: running ``du'' can take a while in case you have a lot of files and
+Note: running \verb|du| can take a while in case you have a lot of files and
 directories; use it with caution!
```
boegel: Maybe we should extend this a bit, by mentioning:

```
Do \epmh{not} (frequently) run `du` on directories where large amounts 
of data are stored, since that will i) take a long time, ii) result in increased
load on the shared storage since (the metadata of) every file in those directories
will have to be inspected.
```

boegel:
OK, good to go, let's bump version to make Travis happy...

PR #239, opened 2018-08-30
Update UGent CAS screenshot

redfast00:


intro-HPC/ch_account.tex
```diff
@@ -351,11 +351,10 @@ \subsection{Users of the \association}
...
 \ifgent
-  You will now have to log in with CAS using your UGent account, and then be requested
-  to share your Mail and Affiliation with KU~Leuven (which actually means the VSC accountpage).
-  Select the ``Remember'' check-box and click ``Yes, continue''.
+  You will now have to log in with CAS using your UGent account, and then you
+  will be requested to share your information. Click ``Yes, continue''.
   \begin{center}
-  \includegraphics*[width=5.73in, height=4.00in, keepaspectratio=false]{ch2-browser-authenticate-gent}
+  \includegraphics*[width=\textwidth]{ch2-browser-authenticate-gent}
   \end{center}
```
boegel: Add some extra information here:

* Your UGent login ID is either a login name of 7 characters (is that fixed?), or a (non-UGent) e-mail address if you are an external user.
* In case of problems with your UGent password, see https://password.ugent.be/

redfast00: It isn't fixed, mine has 8.

boegel: Then use `8 or less`, that's probably right (or good enough in practice) ;)

intro-HPC/ch_account.tex
```diff
@@ -351,11 +351,13 @@ \subsection{Users of the \association}
...
 
 \ifgent
-  You will now have to log in with CAS using your UGent account, and then be requested
-  to share your Mail and Affiliation with KU~Leuven (which actually means the VSC accountpage).
-  Select the ``Remember'' check-box and click ``Yes, continue''.
+  You will now have to log in with CAS using your UGent account. You either have a
+  login name of maximum 8 characters, or a (non-UGent) email address if you are an external
+  user. In case of problems with your UGent password, please visit:
+  \url{https://password.ugent.be/}. After logging in, you will be requested
+  to share your information. Click ``Yes, continue''.
```
boegel: Break this up in paragraph, people lost interest from the 2nd sentence onwards, so important stuff needs to be in the first sentence. ;)

Also, the last sentence is worded too strongly, change `will` to `may be` (you may have approved it already recently)

PR #257, opened 2018-09-04
Remove unused glossary items

### LINKED ISSUE #151: remove unused glossary items from Linux intro
reading the first few pages of the ugent linux docs, there are quite a few gloassary items that seem to be part of the perfexpert docs, but the perferxpert docs seem to be missing from the ugent linux doc. if this is intentional, the corresponding (very technical) items should be removed.

also `last level cache` is abbreviated as `LLC`, not `LCC` (and there's inconsistent usage of `L1` vs `L1d` and friends in the perfexpert text).

redfast00:
Fixes  #151

### DIFF
--- a/glossary.tex
+++ b/glossary.tex
@@ -138,9 +138,9 @@
   name={processor},
   description={A processing unit. A processor is a consumable resource. A processor can be a CPU or a (GP)GPU},
 }
-\newglossaryentry{queues}
+\newglossaryentry{queue}
 {
-  name={queues},
+  name={queue},
   description={PBS/TORQUE queues, or ``classes'' as Moab refers to them, represent groups of computing resources with specific parameters. A queue with a 12 hour runtime or ``walltime'' would allow jobs requesting 12 hours or less to use this queue},
 }
 \newglossaryentry{scp}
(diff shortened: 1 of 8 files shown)


glossary.tex
```diff
@@ -1,31 +1,31 @@
...
-  name={Cluster},
+  name={cluster},
   description={A group of compute nodes},
 }
-\newglossaryentry{Compute-Node}
+\newglossaryentry{compute-node}
 {
-  name={Compute Node},
+  name={compute node},
   description={The computational units on which batch or interactive jobs are processed. A compute node is pretty much comparable to a single personal computer. It contains one or more sockets, each holding a single CPU. Some nodes also contain one or more GPGPUs. The compute node is equipped with memory (RAM) that is accessible by all its CPUs},
```
boegel: We also call this `workernode` or simply `node`, so please add an entry for both and refer to this one.

glossary.tex
```diff
@@ -1,31 +1,31 @@
...
-  name={Distributed memory system},
+  name={distributed memory system},
   description={Computing system consisting of many compute nodes connected by a network, each with their own memory. Accessing memory on a neighbouring node is possible but requires explicit communication},
 }
-\newglossaryentry{Flops}
+\newglossaryentry{FLOPS}
 {
-  name={Flops},
+  name={FLOPS},
   description={Floating-point Operations Per second},
```
boegel: FLOPS is short for "Floating-point Operations Per second", i.e. the number of (floating-point) computations that a processor can perform per second.

glossary.tex
```diff
@@ -88,44 +88,44 @@
...
 {
   name={MPI},
   description={MPI stands for Message-Passing Interface. It supports a parallel programming method designed for distributed memory systems, but can also be used well on shared memory systems},
 }
-\newglossaryentry{Node}
+\newglossaryentry{node}
 {
-  name={Node},
+  name={node},
   description={Typically, a machine, one computer. A node is the fundamental object associated with compute resources},
```
boegel: see also `compute node` (ref to section)

intro-Linux/ch_using_the_clusters.tex
```diff
@@ -1,6 +1,6 @@
 \chapter{Using the clusters}
 
-The use the clusters beyond the login nodes which have limited resources, you
+The use the clusters beyond the \glspl{login-node} which have limited resources, you
```
boegel: add `(s)` so you get `login node(s)`?

PR #272, opened 2018-09-06
Add 'when will my job start?' to FAQ for all sites

### LINKED ISSUE #246: How do users know when their job will start?
See #229: currently, there is a question "When will my job start?" that is wrapped in an `\ifgent`. In Ghent, we don't have the `showstart` command, how about other institutions?

redfast00:
Fixes #246

### DIFF
--- a/intro-HPC/ch_FAQ.tex
+++ b/intro-HPC/ch_FAQ.tex
@@ -1,13 +1,18 @@
 \chapter{Frequently Asked Questions}
 \label{ch:faq}
 
-\ifgent
-% For other institutions: see https://github.com/hpcugent/vsc_user_docs/issues/246
 \section{When will my job start?}
 
-See the explanation about how jobs get prioritized in subsection~\ref{subsec:priority}.
+\ifgent
+See the explanation about how jobs get prioritized in \autoref{subsec:priority}.
+\else
+\ifbrussel
+See the explanation about how jobs get prioritized in \autoref{subsec:priority}.
+\else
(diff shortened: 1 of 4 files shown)


intro-HPC/ch_FAQ.tex
```diff
@@ -1,13 +1,18 @@
...
 
-\ifgent
-% For other institutions: see https://github.com/hpcugent/vsc_user_docs/issues/246
 \section{When will my job start?}
 
-See the explanation about how jobs get prioritized in subsection~\ref{subsec:priority}.
+\ifgent
+See the explanation about how jobs get prioritized in \autoref{subsec:priority}.
+\else
+\ifbrussel
```
boegel: No blocker for this PR, but can you check whether it's feasible to support something like `\ifgent or \ifbrussel`?

redfast00: This is not possible using our current setup.

Ideally, we want to add a macro with a name like `\ifcheckjob` (or whatever property it is Gent and Brussel have in common that makes checkjob unavailable). That way, we don' lose the meaning of the conditional and make it easy to change it if Leuven or Antwerpen decides to disable these commands.

boegel: That's worth considering, but we need to avoid that we start introducing a lot more context-specific macros...

In this case, it makes sense though. `\ifusermoabcommands` maybe makes sense (`checkjob`, `showstart`, etc. are all command that talk to the MOAB job scheduler, and in Ghent & Brussels they're not available to users).

boegel: Can you open a dedicated issue on this to look into it?

Issue #278, opened 2018-09-06
\verb overflows out of the page


redfast00:

redfast00:
`\lstinline` is the solution, now for a global sed across the entire repo

redfast00:
`sed -E 's/verb\|([^|]*)\|/lstinline|\1|/g'`

boegel:
Any references to `\verb` vs `\lstinline`?

redfast00:
@boegel didn't find anything, but I've tested it: `verb` overflows outside of the page, where `lstinline` breaks at spaces if needed.

Closing since PR got merged.

PR #279, opened 2018-09-06
Automatic replacement of verb

redfast00:
See  #278

### DIFF
--- a/run-checks.sh
+++ b/run-checks.sh
@@ -10,7 +10,7 @@ if [[ $result ]]; then
 fi
 
 # Style guide can use \verb to say it's not okay to use \verb
-result=$(grep -R --include=\*.tex '\\verb[^\w]' | grep -v style-guide)
+result=$(grep -R --include=\*.tex '\\verb\W' . | grep -v style-guide)
 if [[ $result ]]; then
     echo "Found a \\verb command, please use \\lstinline instead"
     echo "$result"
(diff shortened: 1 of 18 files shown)


style-guide.tex
```diff
@@ -97,7 +97,7 @@ \section{Displaying example code}
 \section{Displaying inline code}
 \label{sec:displaying-inline-code}
 
-To display inline code inside the document, you can use the \verb|\verb| command.
+To display inline code inside the document, you can use the \lstinline|\lstinline| command.
 This should be used when talking about Linux commands, filenames, \ldots
 When you want inline code inside a title, you can use the \verb|\texttt{}| command.
 When using \verb|\texttt|, you will need to escape certain characters.
```
boegel: Mention why not to use `\verb`?

redfast00:
Checking that `\verb` isn't reintroduced is blocked by #283

run-checks.sh
```diff
@@ -1,15 +1,22 @@
...
 
+result=$(find . -type f -name "*.tex" | xargs -n1 ./find_white_in_terminal.py)
 if [[ $result ]]; then
     echo "Found one or more LaTeX commands in prompt environment, please close them with {}"
     echo "$result"
     ec=1
 fi
 
+# Style guide can use \verb to say it's not okay to use \verb
+result=$(grep -R --include=\*.tex '\\verb[^\w]' | grep -v style-guide)
```
boegel: @redfast00 Usage of `grep` is `grep [OPTIONS] PATTERN [FILE...]`, and `[^\w]` doesn't do what you think it does (I think)?

redfast00:
@boegel I've fixed it and tested it.

Issue #268, opened 2018-09-05
Promt coloring white when using long lines


redfast00:
This is an example:

![v](https://cloud.githubusercontent.com/assets/6941436/3995982/3dc908f8-292f-11e4-90b0-78aade1fa138.png "test ")

(This came originally from issue #71)

redfast00:
This only happens when using macro's. This can be fixed by ending the macro in `{}`, so for example
`authenticity of host %\loginnode% can't be established` becomes `authenticity of host %\loginnode{}% can't be established`.

This probably happens due to the `\xspace` in the `\macro` definition.

boegel:
@redfast00 Can you look into coming up with a `grep` command to detect templates that don't use `{}`?
That way, we can run a script that runs checks like this in Travis, to ensure they're not introduced by accident...

redfast00:
I could write a Python script to detect this, but that wouldn't solve the root cause of this issue: our use of `macro` (more specifically the `\xspace` command in it).

I propose we use the solution in [this TeX StackExchange answer](https://tex.stackexchange.com/a/25825). Here's an example of what that would look like:

```latex
\def\testcommand/{HPC}

\testcommand/ is pretty cool, innit? We can even use it at the end of
a sentence: \testcommand/. Or just at the end of a line \testcommand/

\begin{prompt}
Even in prompts:
%\shellcmd{ssh \testcommand/@\testcommand/}%
%\testcommand/%
\end{prompt}
```

When you forget the trailing `/` in `\testcommand/`, the compiler will error and point you to the line.

This also wouldn't be hard to replace over all documents, since this can be done automatically.

boegel:
So the options are to either add a small check in Travis that fails the tests if a template is used without `{}` when it's needed vs switching to other templates that make the LaTeX build fail.

I see benefits in both, I guess, not sure which one is best... If the error produced by LaTeX is clear (i.e. you don't have to wade through the logs), then that's useful. But it's also more search & replace work.

The downside of doing the check on Travis is that you'll only notice when you open a PR (not when building the PDF locally).

Whatever we go forward with, we should clarify in the style sheet as well.

PR #280, opened 2018-09-07
Fix white in prompt environment

redfast00:
See #268


find_white_in_terminal.py
```diff
@@ -0,0 +1,22 @@
...
+        if '\\begin{prompt}' in line:
+            in_prompt = True
+        elif '\\end{prompt}' in line:
+            in_prompt = False
+        elif '%' in line and in_prompt:
+            # Find latex commands between % %
+            latex_interpreted = re.findall(r'%([^%]*)%', line)
+            for match in latex_interpreted:
+                # Find latex command not ending in {.
+                result = re.search(r'\\[a-zA-Z]+($|[^a-zA-Z{])', match)
```
boegel: Use `\w` rather than `[a-zA-Z]`?

Also, you want to find anything that (starts with `\\` and) doesn't end with `{}`, right?

Would this work? Not sure if allowing whitespace makes sense, is `%  \\example %` valid LaTeX or *must* it be `%\\example%` w.r.t. whitespace?
```python
result = re.search(r'\\\w+\s*$')

File: style-check.sh [hunk 2]
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+result=$(find . -type f -name "*.tex" | xargs -n1 ./find_white_in_terminal.py)
+
+if [[ $result ]]; then
+    echo "Found latex command in prompt environment, please close them with {}"
```
boegel: `Found one or more LaTeX commands in 'prompt' environment, ...`

style-check.sh
```diff
@@ -0,0 +1,12 @@
...
+    echo "$result"
+    exit 1
```
boegel: Don't exit straight away, we'll probably add more checks?

```
ec = 0
...

if [[ $result ]]; then
    ...
    ec=1
fi

exit $ec

File: find_white_in_terminal.py [hunk 4]
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
redfast00: Yes, f-strings (PEP 498).

find_white_in_terminal.py
```diff
@@ -0,0 +1,22 @@
...
+        elif '\\end{prompt}' in line:
+            in_prompt = False
+        elif '%' in line and in_prompt:
+            # Find latex commands between % %
+            latex_interpreted = re.findall(r'%([^%]*)%', line)
+            for match in latex_interpreted:
+                # Find latex command not ending in {.
+                result = re.search(r'\\[a-zA-Z]+($|[^a-zA-Z{])', match)
+                if result is not None:
+                    print(f'{sys.argv[1]} {lineno+1}: {line}')
```
boegel: @redfast00 Let's make this compatible with Python 2.x too, f-strings aren't that much nicer to go only Python 3 imho...

```python
print("%s line %d: %s" % (tex_file, lineno+1, line))

File: find_white_in_terminal.py [hunk 6]
```diff
@@ -5,18 +5,27 @@
...
-                if result is not None:
-                    print(f'{sys.argv[1]} {lineno+1}: {line}')
-                    break
+                # Find latex command not ending in {}.
+                # Uses a negative lookahead (?!...) to assert that the latex
+                #  command doesn't end in a letter or {}
+                for latex_command in re.finditer(r'\\(\w+)(?![\w{])', match):
+                    # Check if we found a latex command not ending in {}
+                    # Allow, since \_ doesn't allow arguments
+                    if latex_command is not None and not latex_command.group(1).startswith('_'):
```
boegel: How can `latex_command` be `None`?

Also, we only want to ignore `_`, not anything that starts with `_` (e.g. `\_foo` should still result in a failed check).

find_white_in_terminal.py
```diff
@@ -0,0 +1,31 @@
...
+        if '\\begin{prompt}' in line:
+            in_prompt = True
+        if '%' in line and in_prompt:
+            # Find content between % %
+            latex_interpreted = re.findall(r'%([^%]*)%', line)
+            for match in latex_interpreted:
+                # Find latex command not ending in {}.
+                # Uses a negative lookahead (?!...) to assert that the latex
+                #  command doesn't end in a letter or {}
+                for latex_command in re.finditer(r'\\(\w+)(?![\w{])', match):
```
boegel: @redfast00 What's the point of looking for `\w` in the negative lookahead? If there's a match for `\w`, it'll be consumed by the `\w+`?
Negative lookahead should be `(?!{})`?

find_white_in_terminal.py
redfast00: `%\_foo%` renders to `_foo` (it doesn't execute the command `_foo`)

find_white_in_terminal.py
redfast00: That was my first idea as well, but that didn't work: consider for example:
`%\shellcmd{whoami}`. The regex `\\(\w+)(?!{})` matches, because `shellcmd` has arguments (and isn't followed by `{}`).

The next logical thing to do would be to use a slightly weaker form: `\\(\w+)(?!{)`. That unfortunately also doesn't work on the example, since it will match `shellcm`, and since the last character (`d`) is not a `{`, it also matches the lookahead.

boegel: OK, thanks for clarifying. :)

Issue #282, opened 2018-09-07
Remove check-version.sh


redfast00:
Replace check-version.sh with a LaTeX command that inserts a timestamp. That way, we won't have merge conflicts all the time.

PR #276, opened 2018-09-06
Specify putty version

redfast00:
Fixes #152

PR #283, opened 2018-09-07
replace check-version.sh with auto-updating datestamp placeholder in Travis config

redfast00:
See #282
screen


intro-HPC/title.tex
```diff
@@ -5,7 +5,7 @@
 \vspace*{1.5\baselineskip}
 
 \Huge \strong{HPC Tutorial} \\
-\LARGE Version 20180906.03
+\LARGE Last updated: \today
```
boegel: @redfast00 This isn't the right way to do it, since then all PDFs will get the same timestamp on deploy, it should be connected to the last made change per document...

boegel:
This looks good to me, it will significantly simplify things since we don't need to worry about the versioning anymore...

I would like to get some feedback on this before merging it though.

Thoughts @wpoely86, @smoors, @mselwa, @sbecuwe, @gjbex?

add-timestamp.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e -u
+
+function set_timestamp() {
+    directory=$1;
+    last_modified=$(git log -n 1 --pretty='format:%cd' --date=format:'%d.%m.%Y' "$1")
```
boegel: @redfast00 I would use `%B %e %Y` as formats which yields `September  6 2018`, rather than `06.09.2018` which leaves it open to interpretation (is it June 9th or Sept 6th?).

boegel:
@redfast00 conflicts will need to be resolved because #276 was merged which did another version update...

wpoely86:
Good for me

smoors:
looks ok to me.
ideally, timestamps would only be updated for the tutorial(s) that have changes (hpc-intro, linux-intro, and/or perfexpert), but it's probably not worth the effort.

boegel:
@smoors That's actually the case now (with the 2nd commit in this PR): the timestamp mentioned in each PDF is based on the date of the last commit in the corresponding directory.

smoors:
ah ok, perfect.

Issue #177, opened 2018-06-26
quota section needs an update


boegel:
For HPC-UGent, it still mentions `show_quota`, which just tells users to visit https://account.vscentrum.be, so we should mention that directly.

What's the correct way to check quota for other sites?

@wpoely86 @hajgato @molden @gjbex @klust

sbecuwe:
Users see their quota when logging in. See \else block in the Pre-defined quotas subsection in ch_running_jobs_with_input_output_data.tex.

mselwa:
for KUL/UH:
quota -s (for $VSC_HOME and $VSC_DATA)
mmlsquota vol_ddn2:leuven_scratch --block-size auto (for $VSC_SCRATCH)

PR #309, opened 2018-09-19
[WIP] Quota

redfast00:
See  #177

### DIFF
--- a/intro-HPC/ch_running_jobs_with_input_output_data.tex
+++ b/intro-HPC/ch_running_jobs_with_input_output_data.tex
@@ -320,23 +320,7 @@ he exceeds the soft quota.
 
 
 \ifgent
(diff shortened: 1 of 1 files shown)


redfast00:
Yes, and the information for Antwerpen doesn't seem to be correct: if I SSH to there, it doesn't display any quota. But then again, it also doesn't work in Leuven, and I don't know what it is in VUB, but if it's `quota -s`, it gives the same error as in Leuven.

I think quota for non-local users might indeed be broken.

boegel:
for VUB: @wpoely86

boegel:
@smoors, @wpoely86: how can VUB users check their quota usage?

smoors:
@boegel quota can be checked with the command `myquota`.

smoors:
it gives a summary of the quota on $VSC_HOME and $VSC_SCRATCH.

PR #313, opened 2018-09-19
Find undefined references

### LINKED ISSUE #295: Travis: detect broken references

redfast00:
Fixes #295


redfast00:
It sure failed, but because time limit exceeded instead of references :/ I'll look into trimming texlive.profile down.

PR #314, opened 2018-09-20
Even less packages

redfast00:


redfast00:
Correct. I'll add in a patch to drop the cache, and remove it as last commit of this PR

boegel:
:+1:

sbecuwe:
No, ConTeXt is not needed for the tutorial.

redfast00:
@sbecuwe we want to trim the texlive install down, so Travis has enough time to install + build everything. Do you know some more things we can remove?

boegel:
@redfast00 Last test on Travis took < 30m (without cache), so trimming things down seems to be helping...

sbecuwe:
It seems latexmk is part of collection-binextra. Sorry!

sbecuwe:
strange: tracklang.sty would be part of collection-plaingeneric, but that is not mentioned in texlive.profile ...
Contents of the different collections can be found in http://mirror.ctan.org/systems/texlive/tlnet/tlpkg/texlive.tlpdb

redfast00:
That URL is really useful! Turns out collection-pstricks depends on collection-plaingeneric.

redfast00:
11 minutes for a full install and compile (so only 5 minutes to install)! I haven't checked if the PDF's look exactly the same, but I'm hopeful.

redfast00:
Please note that this hasn't been tested: it's possible the PDF's look really ugly or are rendered incorrectly: all we know is that they successfully compile, but that's it. If this is merged into master, it would be a good idea to check the pdf's they produced.

boegel:
I'm merging this, I'll take a look at the resulting PDFs and revert these changes if I notice any serious issues...

boegel:
I didn't see any major issues in the resulting PDFs :+1:

PR #327, opened 2018-11-21
update VNC chapter to cover upcoming changes in HPC-UGent login nodes

boegel:
During next week's maintenance, we are switching to new login nodes in the HPC-UGent infrastructure.

This change will me mostly transparent to users, but one important change that is the result of a stricter security policy is that it will no longer be possible to directly connect to a specific login node from outside of the HPC-UGent infrastructure.

Therefore, I have updated the VNC documentation to instruct users to set up the SSH tunnel to `login.hpc.ugent.be` (the generic name for the login nodes), and create an additional tunnel on the login node itself to port forward to the login node where the VNC server is running (if needed).

These instructions are generic, i.e. they will also work even if users are able to directly connect to a specific login node.

### DIFF
--- a/sites/brussel/macros.tex
+++ b/sites/brussel/macros.tex
@@ -10,6 +10,7 @@
 \macro{\hpcname}{Hydra}
 \macro{\loginnode}{hydra.vub.ac.be}
 \macro{\loginhost}{nic50.hydra.vub.ac.be}
+\macro{\altloginhost}{nic51.hydra.vub.ac.be}
 \newcommand{\opensshFirstConnect}{ED25519 key fingerprint is SHA256:m0SKm4laT5OGnGaHIxROJ9LEJJqzsbWB+qpjd4GnOyE
 
 RSA key fingerprint is SHA256:xunjQm9tWulj3hkmSf/3EGs7cOThFIa+Zgn2AZyTtbc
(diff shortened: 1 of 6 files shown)


alvarosimon:
LGTM

sites/brussel/macros.tex
```diff
@@ -10,6 +10,7 @@
 \macro{\hpcname}{Hydra}
 \macro{\loginnode}{hydra.vub.ac.be}
 \macro{\loginhost}{nic50.hydra.vub.ac.be}
+\macro{\altloginhost}{nic51.hydra.vub.ac.be}
```
wpoely86: use `.brussel.vsc` instead of `.vub.ac.be`

sites/gent/macros.tex
```diff
@@ -9,7 +9,8 @@
 \macro{\hpcTeam}{UGent HPC team}
 \macro{\hpcname}{hpcugent}
 \macro{\loginnode}{login.hpc.ugent.be}
-\macro{\loginhost}{gligar01.gligar.gent.vsc}
+\macro{\loginhost}{gligar04.gastly.os}
+\macro{\altloginhost}{gligar05.gastly.os}
```
wpoely86: why use the OS network? Use the VSC names of the machines? Those work from everywhere within the VSC network.

sites/brussel/macros.tex
boegel: Can you open a follow-up PR for that? I just copied what was there already for `\loginhost`...

sites/gent/macros.tex
boegel: Because this corresponds to what they see when they run `hostname` or `hostname -f`

PR #338, opened 2019-03-07
Add "singledocs" for documents that don't depend on os or site.

tdanckaert:
Currently just intro-Cloud.


tdanckaert:
By the way, there's the following comment:

`// filled in from the Makefile by deploy script, don't change them`

Not completely sure what "them" refers to... I just added a new variable for the documents that are independent of OS or institute.  I leave it to you to check if that's correct, or there's some interaction with the Makefile that I missed.

wpoely86:
That comments refers to the `sed` commands in `deploy.sh`. It should be fine for your changes.

PR #339, opened 2019-03-08
Update intro-Cloud

tdanckaert:
Hi,

This should address most of the comments raised on Tuesday, together with a few minor details.

Most substantial changes are in chapter 4 (Configure access and security for instances), and section 5.2 (connect to an instance using SSH). I'd be grateful if you could take a good look at those parts, to see if what I write about ports, firefwalls, SSH and security groups is correct.

Remaining TODO's: requirements on custom images, description of our own images, quota, perhaps a few more glossary items.


tdanckaert:
Can we merge this?

intro-Cloud/ch_ConfigureInstances.tex
```diff
@@ -1,41 +1,73 @@
...
+  ports.  Only the port range 50000-60000 for the floating IP
+  addresses is open by default.  Contact \cloudinfo if you need to
+  access other ports from the outside world.
+\item Instances must connect to the project's \_vm network in order to
+  communicate.
+\item The OpenStack environment has its own internal firewalls, which
+  block most ports of your instances.  If you want to access specific
+  ports of your instances, you must create ``Security Groups'' which
+  allow access to those ports.
+\item If you want to access your instances using SSH key
```
alvarosimon: Just change this item a bit. Users do not have to upload their keys because this is sync with the account portal, so they only have to upload their keys to the account portal as they do for HPC (the sync script will update the user's openstack keys automatically from there). Maybe just include the link from the hpc doc to know how to upload keys to the vsc system (section 2.2 I think)

alvarosimon:
@tdanckaert looks nice, just one small remark

alvarosimon:
Also we should include an example as Alexander said about the host ssh offending key.

They can use `ssh-keygen -R [project_public_ip]:<port>` to remove the offending key automatically:
https://askubuntu.com/questions/455635/how-to-remove-offending-key-in-git-when-you-try-to-push-your-changes
They should also check that the host key does match with the key provided by the openstack VM (just taking a look in the dashboard the VM log tab).

This is quite common if the user re-use the same port and public IP with different VMs or if the VM is re-instantiated.

intro-Cloud/ch_ConfigureInstances.tex
tdanckaert: This section does not talk about uploading new keys.  You still need to ask OpenStack to inject your key to instances you create, regardless of how that key was imported into OpenStack (by the user themselves or by the synchronization with the account portal).  I've added a sentence mentioning the account keys to make it more clear, hopefully (all of this discussed in more detail in other sections, this is just a general introduction to the chapter).

tdanckaert:
I've added 2 pages about host keys and related warnings and errors.  Currently, this section is specific to OpenSSH.  Do we have (or expect) cloud users working with putty from Windows?

intro-Cloud/ch_ConfigureInstances.tex
alvarosimon: ok, in that case I would include a reference to the specific ssh section here, just to avoid any confusion (something like `for more info about ssh keys injection see \ref{sec:ssh-keypairs}`). In that case they can go directly to the detailed info.

tdanckaert: Ok, I've shortened the introduction (no need to duplicate information here) and added references to the relevant sections.  I've also reordered the sections to match the order of in which they are introduced, instead of the arbitrary order in which they appeared until now.

alvarosimon:
> I've added 2 pages about host keys and related warnings and errors. Currently, this section is specific to OpenSSH. Do we have (or expect) cloud users working with putty from Windows?

yes, and they should use putty+pageant to access to the VMs. Fortunately this information is already available from the HPC doc for windows users. We should include the link with a note for Windows users, if I'm not mistaken the info is available in sections `2.1.2` and `2.1.4` from intro-HPC-windows

tdanckaert:
> yes, and they should use putty+pageant to access to the VMs. Fortunately this information is already available from the HPC doc for windows users. We should include the link with a note for Windows users, if I'm not mistaken the info is available in sections `2.1.2` and `2.1.4` from intro-HPC-windows

Ah it's in sections 8.4 and 8.6 (I didn't know!).  I'll point windows users to those sections.

alvarosimon:
@tdanckaert are you going to add or change something else in the current PR?

tdanckaert:
No, you can merge it.

alvarosimon:
@tdanckaert ok thanks a lot!

PR #343, opened 2019-04-29
remove section discussing submitting to specific job queues

boegel:
I propose we drop the section explains submitting to specific queues entirely, since we agreed a long time ago that users shouldn't be submitting to specific queues but just request walltime/node properties/etc. instead.
Which queues are defined is pretty much irrelevant (and also different among the different VSC sites, so very confusing as well).

Large parts of this section were already disabled using `\iffalse`, only the part until `\inputsite{queue-table}` was still active.

@klust @sbecuwe @wpoely86 @smoors @mselwa @gjbex Since this also implies removing the table that mentions queue limits in terms of job count, some of this information should probably be reworded in terms of requested walltime rather than queue names?


intro-HPC/ch_MATLAB.tex
```diff
@@ -105,6 +105,14 @@ \subsection{Memory issues during compilation}
 %\shellcmd{export \_JAVA\_OPTIONS="-Xmx512M"}%
 \end{prompt}
 
+\subsection{Figures}
+
+figure()
+
+wordt
```
wpoely86: english?

boegel: Ah, crap, that shouldn't have been in here... Will remove & force push.

wpoely86:
Fine for me to drop. The only limits the user should care about are the walltime limits.

stdweird:
to be very clear, once this is removed from the docs, we will remove queues on our site altogether, and requersting any queue will fail hard.

boegel:
@klust @sbecuwe @wpoely86 @smoors @mselwa @gjbex ping on making this change?

smoors:
for us this is ok, queues will be handled by our submitfilter that is soon-to-go-in-production.

sbecuwe:
ok, queues are still active in Antwerp, but we only talk about wall time during the intro session.

boegel:
@mselwa @gjbex Any objections from KUL on getting this merged? If we don't hear back, we'll get this in by the end of next week.

PR #345, opened 2019-05-03
Notify users about forced fall back of non-supported locale. HPC-7246

hajgato:


intro-HPC/ch_connecting.tex
```diff
@@ -504,6 +504,13 @@ \section{First Time connection to the HPC infrastucture}
 user interface. Usually a locale identifier consists of at least a language
 identifier and a region identifier.
 
+\ifgent
+\strog{Note:} If you try to set on a UGent HPC cluster a non-supported locale,
```
stdweird: remove the `on a UGent HPC cluster`, this is a `\ifgent` block

hajgato: Done.

PR #346, opened 2019-05-15
fix documentation w.r.t. checking quota and getting more quota for HPC-UGent VSC accounts

boegel:
There's little point in letting HPC-UGent users run `show_quota` on the HPC-UGent login nodes, since all it does is print this:

```
$ show_quota
Please consult the VSC account page for quota information at
https://account.vscentrum.be

If you are a VO moderator, you can find the VO quota for all members at
https://account.vscentrum.be/django/vo/
```

### DIFF
--- a/intro-HPC/ch_running_jobs_with_input_output_data.tex
+++ b/intro-HPC/ch_running_jobs_with_input_output_data.tex
@@ -454,12 +454,12 @@ The available disk space on the \hpc is limited. The actual disk capacity,
 shared by all users, can be found on the ``Available hardware'' page on the
 website.
 (\url{https://www.vscentrum.be/infrastructure/hardware})
-As explained in \S\ref{subsec:predfined-quotas}, this implies
-that there are also limits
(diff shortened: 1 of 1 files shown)


intro-HPC/ch_running_jobs_with_input_output_data.tex
```diff
@@ -515,6 +523,19 @@ \subsection{Check your quota}
 % Scratch        11  100000  150000       20   none
 % \end{prompt}
 
+\ifgent
+If you obtained your VSC account via \university, you can consult your current storage quota usage
```
alvarosimon: I would replace this sentence by just:
`You can consult your current \university storage quota...`  if we use `\ifgent` we assume that this documentation is for UGent users and you do not need to copy/paste the previous tip.

boegel: @alvarosimon fixed in f24bbe9

PR #357, opened 2019-09-19
enable pipefail mode to catch 'make all' failures

boegel:
This ensures that the exit code of `make all 2>&1 | tee $LOGFILE` will be non-zero if `make all` fails.


boegel:
works as intended, see https://travis-ci.org/boegel/vsc_user_docs/builds/587009510 (which was run without the fix from #356)

PR #364, opened 2019-11-18
Add remark in running_batch_jobs to clearify you should reload modules in your job script

### LINKED ISSUE #363: Make clear how to load modules for a job
In section _4.3 Defining and submitting your job_ it is not clear how you should use modules for your job.
The 'logical' way would be to `module load` all your modules then submit a job and everything should work. It is not mentioned that this is not the case and you should `module load` all your required modules in the batch job script.

ajuvercr:
Fix #363

### DIFF
--- a/intro-HPC/ch_running_batch_jobs.tex
+++ b/intro-HPC/ch_running_batch_jobs.tex
@@ -720,6 +720,10 @@ The qsub command returns a job identifier on the HPC cluster. The important
 part is the number (e.g., ``\jobnumber''); this is a unique identifier for the job
 and can be used to monitor and manage your job.
 
+\underbar{Remark}: the modules that were loaded when you submitted the job won't be loaded
+when the job is started. So you should reload them in your jobscript. This can be
+very tedious but you can use the before mentioned module collections.
+
 Your job is now waiting in the queue for a free workernode to start on.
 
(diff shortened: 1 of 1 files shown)


intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -720,6 +720,10 @@ \section{Defining and submitting your job}
 part is the number (e.g., ``\jobnumber''); this is a unique identifier for the job
 and can be used to monitor and manage your job.
 
+\underbar{Remark}: the modules that were loaded when you submitted the job won't be loaded
+when the job is started. So you should reload them in your jobscript. This can be
+very tedious but you can use the before mentioned module collections.
```
boegel: `won't be loaded`  -> `will \emph{not} be loaded`

`So you should reload them in your jobscript` -> `You should always specify the \lstinline|module load| statements that are required for your job in the job script itself.`

`This can be very tedious but` ->
```
\ifusinglmod
To faciliate this, you can use a pre-defined module collection which you can restore using \lstinline|module restore|.
\fi
```

wdpypere:
@boegel I don't presume @ajuvercr will make anymore changes to this. Can you have a look?

boegel:
@wdpypere tackled via #378, merging that PR will result in merging this one too

PR #367, opened 2019-12-05
256 -> 255 in ssh-ed25519 host key for login nodes

boegel:
Recent PuTTy versions show `ssh-ed25519 255` rather than `ssh-ed25519 256`, which confuses users.

See also screenshot:
image


alvarosimon:
This is a known issue in the last putty client version. The fingerprint size is incorrect (255 instead 256). This should be fixed in a new putty update. In any case to avoid to change this in the future we can just add a note for the users. The real value that they should check is the ssh fingerprint string, not the size.

wdpypere:
@boegel @alvarosimon So does something need to be added to the documentation to explain the bug?
Can this be closed?

boegel:
@alvarosimon Added the note you suggested should be OK to merge now...

PR #387, opened 2020-08-04
run kdestroy on the end of session

kwaegema:


intro-HPC/ch_running_jobs_with_input_output_data.tex
```diff
@@ -364,6 +364,14 @@ \subsection{Your UGent home drive and shares \label{subsec:ugent-shares}}
 
 Each hour the process will check if your ticket should be renewed.
 
+
+We advise to disable access to your shares once access is no longer needed:
+
+\begin{prompt}
+%\shellcmd{kdestroy}%
```
wdpypere, line 371: I would use `kdestroy -A`

stdweird: i thought about that, but it will also destroy any other tokens (however unlikely)
i would only destroyt the tokens created (if possible).

boegel: So which token will `kdestroy` destroy then? Does it ask, or what?

itkovian: ALL OF THEM.

intro-HPC/ch_running_jobs_with_input_output_data.tex
```diff
@@ -364,6 +364,14 @@ \subsection{Your UGent home drive and shares \label{subsec:ugent-shares}}
 
 Each hour the process will check if your ticket should be renewed.
 
+
+We advise to disable access to your shares once access is no longer needed:
```
itkovian, line 368:
```suggestion
We strongly advise to disable access to your shares once it is no longer needed:
```

intro-HPC/ch_running_jobs_with_input_output_data.tex
wdpypere: you can either:
 - destroy all caches of the user (-A)
 - destroy a specific cache (-c)
 - destroy the default cache (no flags)

in practice, I doubt a user will use anything else than the default cache location.

stdweird: ah, it destroys the whole cache, not individual tokens.
it looks like `klist purge` is someting one could use for that, but too complicated to explain and not worth it. `kdestroy` is more than enough

PR #390, opened 2020-09-01
delete PBS_O_(V)NODENUM variables

hajgato:
We do not have in Gent, but I have checked in Brussels, and PBS_O_NODENUM and PBS_O_VNODENUM variables are also missing. Anyway, I do not understand what did this mean based on the explanation. (Google did not help either, but I barely made any effort)


stdweird:
shouldn't we add this to the wrappers?

wpoely86:
Triggering actions test on this PR.

PR #398, opened 2020-11-01
UA update October 2020

sbecuwe:
- replaced \include by \input in glossary.tex to avoid problems after October update of LaTeX
- updated macros for UAntwerpen: new fingerprints + replaced hopper by leibniz
- update Account page figure (new logo) and UAntwerpen idp in section 2.2


intro-HPC/intro-HPC.tex
```diff
@@ -67,7 +67,10 @@ \part{Beginner's Guide}
 
 \include{ch_multi_core_jobs}
 
+\ifantwerpen
+\else
 \include{ch_web_portal}
```
boegel, line 72: @sbecuwe This should probably be under `\ifgent` instead?

intro-HPC/title.tex
```diff
@@ -81,7 +81,9 @@
 How to start jobs with user interaction?                                & \strong{\ref{ch:running-interactive-jobs}} & \nameref{ch:running-interactive-jobs} \\ \hline
 Where do the input and output go? Where to collect my results?          & \strong{\ref{ch:running-jobs-with-input-output-data}} & \nameref{ch:running-jobs-with-input-output-data} \\ \hline
 Can I speed up my program by exploring parallel programming techniques? & \strong{\ref{ch:multi-core-jobs-parallel-computing}} & \nameref{ch:multi-core-jobs-parallel-computing} \\ \hline
+\ifthenelse{\not \boolean{antwerpen}}{
```
boegel, line 84: Same here, this should be under `\ifgent`, we overlooked that...

boegel (changes requested):

intro-HPC/title.tex
sbecuwe: Since neither colleagues from Leuven or Brussels removed it from their version, I thought it should stay there.
So no objection removing it. I will update both files.

boegel (approved):

PR #416, opened 2021-08-02
Why my job does not start and there are free nodes?

hajgato:


intro-HPC/ch_running_batch_jobs.tex
```diff
@@ -795,6 +795,12 @@ \subsection{When will my job start?}
 in a workernode), since the scheduler often finds a ``gap'' to fit the job into
 more easily.
 
+Sometimes it happens that couple of nodes are free and a job would not start.
+Empty nodes are not necessary empty for your job(s). Just imagine, that an
+\emph{N}-node-job should run. It is quite unlikely that \emph{N} nodes would be the same moment empty,
+so until \emph{N-1} nodes become empty, you will see them, and the moment the \emph{N}th node
+becomes empty the \emph{N}-node-job will accommodate the \emph{N} free nodes.
```
itkovian, line 802:
```suggestion
\emph{N}-node-job (with a higher priority than your waiting job(s)) should run. It is quite unlikely that \emph{N} nodes would be empty at the same moment to accommodate this job, so while fewer than \emph{N} nodes are empty, you can see them as being empty. The moment the \emph{N}th node becomes empty the waiting \emph{N}-node-job will consume these \emph{N} free nodes.
```

hajgato: @itkovian Done

PR #464, opened 2022-11-17
fix URL of VSC firewall app

boegel:
used to be https://firewall.hpc.kuleuven.be, now https://firewall.vscentrum.be

minor issue, since https://firewall.hpc.kuleuven.be automatically redirects to https://firewall.vscentrum.be


alvarosimon:
Fixed by #506

PR #493, opened 2023-04-14
Fix build script

stdweird:


boegel:
@stdweird Please add some info on the use of `$INSTALLDEPS` and `$WEBSERVER` to control `build.sh` to the README?

Or maybe it's better to use proper CLI options for the script, so you can do `./build.sh --install-deps --preview` or whatever?

build.sh
```diff
@@ -1,5 +1,28 @@
 #!/bin/bash
-# python3 -m pip install --ignore-installed --prefix $PWD/pypkgs -r requirements.txt
-# python3 -m pip install --prefix $PWD/pypkgs $PWD/custom_plugin
-# python3 -m pip install --prefix $PWD/pypkgs $PWD/computational_macros
-PYTHONPATH=$PWD/pypkgs/lib/python3.9/site-packages:$PWD/custom_plugin:$PWD/computational_macros:$PYTHONPATH python3 build.py
+set -x
+if [ ${INSTALLDEPS:-0} -gt 0 ]; then
+    python3 -m pip install --ignore-installed --prefix $PWD/pypkgs -r requirements.txt
```
alvarosimon, line 4: `shellcheck build.sh`  shows several warnings such using  double quotes for all variables like "$PWD"

stdweird:
@boegel don't know what is best. if we add proper cli options, we need proper --help; and now i just pass the cli opts to the build.py
i'll add them to the README for now

stdweird:
@boegel i'm assuming that the mkdocs commands are independent

PR #503, opened 2023-04-20
Update mkdocs-material and other requirements, and add underscore as search separator

### LINKED ISSUE #501: search does not find words in code part of the docs
docs do not search in codes, for example no ARCANINE found at all, but it is in the docs: https://docs.hpc.ugent.be/Linux/running_jobs_with_input_output_data/#where-to-store-your-data-on-the-hpc

stdweird:
Fixes #501


stdweird:
@boegel @alvarosimon if you try this, you must remove  local deps (`rm -Rf pypkgs`) first, or you will e.g. miss the search box

alvarosimon:
@stdweird does this `requirements.txt` work for you? because I have tried these new versions and the search box just disappears, testing this from my computer (probably I'm missing something)

alvarosimon:
@stdweird ok I didn't read your previous comment :-) I will remove local deps first

PR #504, opened 2023-04-21
Add search share in cloud

alvarosimon:
I have included and tested this just for cloud, it is a nice feature for otrs.

### DIFF
--- a/config/others/cloud.yml
+++ b/config/others/cloud.yml
@@ -57,18 +57,15 @@ theme:
   name: material
   language: en
   features:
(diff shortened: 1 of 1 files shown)


stdweird:
@alvarosimon i have merged this, now @boegel and @hajgato can look if this is also useful for hpc docs

alvarosimon:
@stdweird ok for cloud is quite useful if you want to share with a user a topic that is included in several sections to avoid copy/pasting all the links.

PR #532, opened 2023-08-07
Overview page modules

milachae:


boegel (changes requested):

scripts/module_list/README.md
```diff
@@ -0,0 +1,2 @@
+# Module list
+A script that generates a list (in markdown) of all available lmod modules.
```
boegel, line 2: Add basic usage info + list of requirements (Lmod must be available, `$LMOD_CMD` must specify path to `lmod` binary, `$MODULEPATH` must be set up correctly, required Python packages outside of Python stdlib lib (see `requirements.txt`), etc.)

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
+import os
+import subprocess
+from typing import Union, Tuple
+from mdutils.mdutils import MdUtils
+import numpy as np
```
boegel, line 5: Add license header, see for example https://github.com/hpcugent/vsc-base/blob/master/lib/vsc/utils/run.py

and order imports alphabetically:

```python
import numpy as np
import os
import subprocess
from mdutils.mdutils import MdUtils
from typing import Union, Tuple

File: scripts/module_list/module_overview.py [hunk 3]
```diff
@@ -0,0 +1,175 @@
...
+
+
+# --------------------------------------------------------------------------------------------------------
+# Module bash API
+# --------------------------------------------------------------------------------------------------------
+
+def module(*args, filter_fn=lambda x: x) -> np.ndarray:
+    """
+    API to call the module command.
```
boegel, line 14:
```suggestion
    Function to run Lmod commands, via Python interface.
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @param args: Extra arguments for the module command.
+    @param filter_fn: Filter function on the ouput.
+    @return:
+    """
+    lmod = os.getenv('LMOD_CMD')
```
boegel, line 20: No need to implement in this PR, but we should add some error handling here.
What if `$LMOD_CMD` is not set (or points to a non-existing path)?
What if the exit code of the Lmod command is non-zero?

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+    proc = subprocess.run(
+        [lmod, "python", "-t"] + list(args),
+        encoding="utf-8",
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE
+    )
+    exec(proc.stdout)
+    return filter_fn(np.array(proc.stderr.split()))
```
boegel, line 28: Maybe add a comment to explain what is contained in `stdout` (Python statement to change current environment) and `stderr` (actual output produced by module command, in machine-readable format because we use `--terse`)

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    API to call the module command.
+
+    @param args: Extra arguments for the module command.
+    @param filter_fn: Filter function on the ouput.
+    @return:
+    """
+    lmod = os.getenv('LMOD_CMD')
+    proc = subprocess.run(
+        [lmod, "python", "-t"] + list(args),
```
boegel, line 22: use `--terse` instead of `-t`, to make this "self-documenting", `-t` could be anything

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE
+    )
+    exec(proc.stdout)
+    return filter_fn(np.array(proc.stderr.split()))
+
+
+def module_avail(name: str = "", filter_fn=lambda x: x) -> np.ndarray:
+    """
+    API to call the module avail command of Lmod.
```
boegel, line 33:
```
    Function to run "module avail" commands.
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @param name: Possible module name.
+    @param filter_fn: Filter on the output.
+    @return: List of all available modules of name, or all if name is not given.
+    """
+    # return module(f"avail {name if name else ''}", filter_fn=filter_fn)
```
boegel, line 39: remove commented out code?

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+# Module bash API
+# --------------------------------------------------------------------------------------------------------
+
+def module(*args, filter_fn=lambda x: x) -> np.ndarray:
+    """
+    API to call the module command.
+
+    @param args: Extra arguments for the module command.
+    @param filter_fn: Filter function on the ouput.
+    @return:
```
boegel, line 18: missing info on return value?

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
+import os
+import subprocess
+from typing import Union, Tuple
+from mdutils.mdutils import MdUtils
+import numpy as np
+
+
+# --------------------------------------------------------------------------------------------------------
+# Module bash API
```
boegel, line 9:
```suggestion
# Functions to run "module" commands
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,175 @@
...
+    structured, col, row = generate_table_data(data)
+    md_file.new_table(columns=col, rows=row, text=list(structured), text_align='center')
+
+
+def generate_general_overview() -> None:
+    """
+    Generate the general overview in a markdown file.
+    It generates a list of all the available software and indicates on which cluster it is available.
+    """
+    md_file = MdUtils(file_name='Example_Markdown', title='Overview Moduls')
```
boegel, line 167:
```suggestion
md_file = MdUtils(file_name='module_overview.md', title='Overview of available modules per cluster')

**Review by boegel — CHANGES_REQUESTED** [review]

File: scripts/module_list/tests/test_module.py [hunk 12]
```diff
@@ -0,0 +1,24 @@
...
+    @classmethod
+    def setup_class(cls):
+        os.environ["LMOD_CMD"] = cls.path + "data/lmod_mock.sh"
+
+    # ---------------------------
+    # Module tests
+    # ---------------------------
+
+    def test_avail(self):
+        os.environ["MOCK_FILE"] = self.path + "data/mock_avail_0.txt"
```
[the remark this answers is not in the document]
milachae: I probably forgot to change this file. This is fixed now.

scripts/module_list/tests/data/data_avail_simple.txt
```diff
@@ -0,0 +1,29 @@
+/apps/gent/RHEL8/zen2-ib/modules/all:
```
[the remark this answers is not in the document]
milachae: Changed it to made up data.

> If you have two clusters in the mocked output for module avail cluster, then you should also have two files that contain the matching fake output for module avail for those clusters.
Note that this should always include the cluster modules too, see:

>For now, you can pretend like those cluster modules are not there (so they'll end up in the overview you generate), but in a later stage you should also implement a filtering mechanism, where modules in a particular subdirectory (like /etc/modulefiles/vsc) can be filtered out so they're not included in the overview.

These are already included, but are filtered out by a filter function. The default function is `lambda x: x`, but I made special filters for the output of the clusters at UGent.

scripts/module_list/tests/test_data.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_clusters_ugent(self):
+        os.environ["MOCK_FILE_AVAIL_CLUSTER"] = self.path + "/data/doduo_avail_cluster.txt"
+        sol = clusters_ugent()
+        assert len(sol) == 8
+        assert list(sol) == ['cluster/accelgor', 'cluster/doduo', 'cluster/donphan',
+                             'cluster/gallade', 'cluster/joltik', 'cluster/skitty',
+                             'cluster/swalot', 'cluster/victini']
+
+    def test_large(self):
+        os.environ["MOCK_FILE_AVAIL"] = self.path + "/data/doduo_avail.txt"
```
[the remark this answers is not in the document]
milachae: I deleted these files for now.

scripts/module_list/tests/test_data.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_data_ugent(self):
+        os.environ["MOCK_FILE_AVAIL"] = self.path + "/data/data_avail_simple.txt"
+        os.environ["MOCK_FILE_AVAIL_CLUSTER"] = self.path + "/data/data_avail_cluster_simple.txt"
+        sol = modules_ugent()
+        assert len(sol) == 2
+        assert len(sol["cluster/victini"]) == 15
+        assert list(simplify_modules(sol["cluster/victini"])) == ["ABAQUS", "ABINIT", "zstd"]
+
+    def test_clusters_ugent(self):
+        os.environ["MOCK_FILE_AVAIL_CLUSTER"] = self.path + "/data/doduo_avail_cluster.txt"
```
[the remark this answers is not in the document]
milachae: This file does not exists anymore. Large files could be added later.

.github/workflows/script_module_list.yml
```diff
@@ -0,0 +1,41 @@
...
+          path: "scripts/module_list"
+
+  pytest-tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
```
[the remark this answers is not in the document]
milachae: This has been fixed now.

scripts/module_list/README.md
milachae: README.md has been updated.

scripts/module_list/module_overview.py
milachae: done

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+# https://github.com/hpcugent/vsc_user_docs
+#
+# vsc_user_docs is licensed under a
+# Creative Commons Attribution-ShareAlike 4.0 International License.
+#
+# You should have received a copy of the license along with this
+# work. If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
+#
+"""
+Python script to convert all available modules in lmod to a markdown list.
```
boegel, line 20:
```suggestion
Python script to generate an overview of available modules across different clusters, in MarkDown format.
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+    exec(proc.stdout)
+    return filter_fn(np.array(proc.stderr.split()))
+
+
+def module_avail(name: str = "", filter_fn=lambda x: x) -> np.ndarray:
+    """
+    Function to run "module avail" commands.
+
+    @param name: Possible module name.
```
boegel, line 59:
```suggestion
    @param name: Module name, or empty string to return all available modules.

File: scripts/module_list/module_overview.py [hunk 19]
```diff
@@ -0,0 +1,201 @@
...
+    """
+    return data[~np.char.endswith(data, ":") &
+                ~np.char.startswith(data, "env/") &
+                ~np.char.startswith(data, "cluster/default")
+                ]
+
+
+def filter_fn_gent_modules(data: np.ndarray) -> np.ndarray:
+    """
+    Filter function for the output of all modules.
```
boegel, line 96:
```suggestion
    Filter function for the output of all software modules (excl. `cluster` and `env` modules).
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+    Returns all the cluster names of the HPC at UGent.
+    @return: cluster names
+    """
+
+    return module_avail(name="cluster/", filter_fn=filter_fn_gent_cluster)
+
+
+def modules_ugent() -> dict:
+    """
+    Returns all the module names that are installed on the HPC on UGent.
```
boegel, line 117:
```suggestion
    Returns names of all software module that are installed on the HPC on UGent.

File: scripts/module_list/module_overview.py [hunk 21]
```diff
@@ -0,0 +1,201 @@
...
+        module_swap(cluster)
+        data[cluster] = module_avail(filter_fn=filter_fn_gent_modules)
+    return data
+
+
+# --------------------------------------------------------------------------------------------------------
+# Util functions
+# --------------------------------------------------------------------------------------------------------
+
+def simplify_modules(data: Union[dict, list, np.ndarray]) -> Union[dict, list, np.ndarray]:
```
boegel, line 133: Maybe rename to `get_software_names`?

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Simplify list of modules by removing versions and duplicates.
+
+    @param data: List of modules
+    @return: List of programs.
```
boegel, line 138: `programs` -> `software names`

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+
+    if isinstance(data, dict):
+        simplified_data = {}
+        for cluster in data:
+            simplified_data[cluster] = np.unique([entry.split("/")[0] for entry in data[cluster]])
+    else:
+        simplified_data = np.unique([entry.split("/")[0] for entry in data])
```
boegel, line 146: use a helper function:

```python
    def mod_names_to_software_names(mod_names):
        return np.unique([m.split('/')[0] for m in mod_names])

    if isinstance(data, dict):
        simplified_data = dict((x, mod_names_to_software_names(data[x])) for x in data)
    else:
        simplified_data = mod_names_to_software_names(data)

File: scripts/module_list/module_overview.py [hunk 24]
```diff
@@ -0,0 +1,201 @@
...
+
+    return simplified_data
+
+
+# --------------------------------------------------------------------------------------------------------
+# Generate markdown
+# --------------------------------------------------------------------------------------------------------
+
+def generate_table_data(data: dict) -> Tuple[np.ndarray, int, int]:
```
boegel, line 155: maybe use a more meaningful name here than `data`, since this data as a particular structure; how about `avail_mods`?

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+    Generate data that can be used to construct a MarkDown table.
+
+    @param data: Available data
+    @return: Returns tuple (Table data, #col, #row)
+    """
+    data = simplify_modules(data)
+    all_modules = simplify_modules(np.concatenate(list(data.values())))
+
+    final = np.array([" "])
+    final = np.append(final, list(data.keys()))
```
boegel, line 166: make sure column names don't have `cluster/` in them, so extract a list of cluster names first, something like:

```python
cluster_names = [x.lstrip('cluster/') for x in data.keys()]
final = np.append(final, cluster_names)
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+
+    for package in all_modules:
+        final = np.append(final, package)
+
+        for cluster in data:
+            final = np.append(final, "X" if package in data[cluster] else " ")
+
+    return final, len(data.keys()) + 1, len(all_modules) + 1
```
boegel, line 174:
```suggestion
    return final, len(cluster_names) + 1, len(all_modules) + 1
```

scripts/module_list/module_overview.py
```diff
@@ -0,0 +1,201 @@
...
+def modules_ugent() -> dict:
+    """
+    Returns all the module names that are installed on the HPC on UGent.
+    They are grouped by cluster.
+    @return: Dictionary with all the modules per cluster
+    """
+
+    data = {}
+    for cluster in clusters_ugent():
+        module_swap(cluster)
```
boegel, line 124: Some output should be generated here, since this takes a while. Something like:

```
Collecting available modules for cluster/doduo... found 1234 modules!
Collecting available modules for cluster/donphan...
...
```

scripts/module_list/requirements.txt
```diff
@@ -0,0 +1,5 @@
+flake8
+pytest
+mdutils
+numpy
+setuptools
```
boegel, line 5: Why is `setuptools` an explicit dependency?

I guess only `mdutils` and `numpy` really are, and maybe there should be a separate `requirements_tests.txt` for `flake8` and `pytest`?

scripts/module_list/test.sh
```diff
@@ -0,0 +1 @@
+PYTHONPATH=$PWD:$PYTHONPATH pytest -v -s
```
boegel, line 1: Add `#!/bin/bash` on top

scripts/module_list/tests/data/lmod_mock.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Emulated swap command.
+elif [ "$3" = "swap" ]; then
+  #extract te cluster name from the 4th argument
+  cluster=$(echo "$4" | cut -d "/" -f 1)
+  cluster_name=$(echo "$4" | cut -d "/" -f 2)
+
+  if [ "$cluster" = "cluster" ]; then
+    # Substitute CLUSTER by the cluster_name
+    echo "${MOCK_FILE_SWAP}" | sed "s/CLUSTER/${cluster_name}/" | xargs cat >&1
```
boegel, line 20: This works too I think?
```
cat ${MOCK_FILE_SWAP/CLUSTER/${cluster_name}}
```

scripts/module_list/tests/data/lmod_mock.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
boegel, line 2: add `set -u` the produce an error when the expected environment variable is not defined?

scripts/module_list/tests/test_md.py
```diff
@@ -0,0 +1,46 @@
...
+    # Markdown tests
+    # ---------------------------
+
+    def test_table_generate_simple(self):
+        simple_data = simplify_modules(modules_ugent())
+        table_data, col, row = generate_table_data(simple_data)
+        all_modules = simplify_modules(np.concatenate(list(simple_data.values())))
+        assert col == len(simple_data.keys())+1
+        assert row == len(all_modules)+1
+        assert len(table_data) == (len(simple_data.keys()) * len(all_modules)) + \
```
boegel, line 36: Why not use hard counts here, like in `test_data_ugent`?

scripts/module_list/tests/test_md.py
```diff
@@ -0,0 +1,46 @@
...
+               len(simple_data.keys()) + \
+               len(all_modules) + \
+               1
+
+    def test_simple(self):
+        md_file = MdUtils(file_name='test_simple', title='Overview Modules')
+        simple_data = simplify_modules(modules_ugent())
+        generate_module_table(simple_data, md_file)
+        md_file.create_md_file()
+        assert os.path.exists("test_simple.md")
```
boegel, line 46: Maybe check first 3 lines or so?
Or check specific lines, like the `llm` one which is there on one fake cluster, not another.

In fact, you can probably include a constant with the expected output, and hard check against that:

```python
EXPECTED_MD_TABLE = """...
"""
...
    def test_simple(self):
       ...
       with open('test_simple.md') as fp:
           assert fp.read() == EXPECTED_MD_TABLE

File: scripts/module_list/tests/data/lmod_mock.sh [hunk 34]
```diff
@@ -0,0 +1,30 @@
...
+  if [ "$4" = "cluster/" ]; then
+    cat "${MOCK_FILE_AVAIL_CLUSTER}" >&2
+  else
+    cat "${MOCK_FILE_AVAIL}" >&2
+  fi
+
+
+# Emulated swap command.
+elif [ "$3" = "swap" ]; then
+  #extract te cluster name from the 4th argument
```
boegel, line 14:
```suggestion
  # extract the cluster name from the 4th argument
```

scripts/module_list/tests/data/lmod_mock.sh
```diff
@@ -0,0 +1,30 @@
...
+  cluster=$(echo "$4" | cut -d "/" -f 1)
+  cluster_name=$(echo "$4" | cut -d "/" -f 2)
+
+  if [ "$cluster" = "cluster" ]; then
+    # Substitute CLUSTER by the cluster_name
+    echo "${MOCK_FILE_SWAP}" | sed "s/CLUSTER/${cluster_name}/" | xargs cat >&1
+  else
+    echo "Something is wrong"
```
boegel, line 22: explain *what* is wrong ("`Something is wrong: expected 'cluster', got '${cluster}'`")

and use `>&2` to redirect the message to stderr

scripts/module_list/tests/data/lmod_mock.sh
boegel, line 2: name the args you get (even if you don't use them below)
```shell
lmod=$1
lmod_shell=$2
mod_cmd=$3
mod_args=$4

File: scripts/module_list/tests/data/lmod_mock.sh [hunk 36]
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+  fi
+
+
+else
+  echo "Not supported"
```
boegel, line 28:
```suggestion
    echo "Module subcommand '${mod_cmd}' not supported yet in $0" >&2

**Review by boegel — CHANGES_REQUESTED** [review]

File: scripts/module_list/README.md [hunk 37]
```diff
@@ -0,0 +1,79 @@
...
+3. **Utilizing the Swap Command:**
+
+   For utilizing the swap command, assign the path to the swap files to the MOCK_FILE_SWAP variable. 
+   Ensure that the filename contains the placeholder CLUSTER, 
+   which will later be replaced with the actual cluster name when performing the swap.
+
+   ```python
+   os.environ["MOCK_FILE_SWAP"] = path + "/data/data_swap_CLUSTER.txt"
+   ```
+   When trying to swap to, for example, the cluster/dialga cluster.
```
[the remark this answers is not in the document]
milachae: :( no pikachu

scripts/module_list/tests/data/lmod_mock.sh
milachae: Some env variables are intentionally not defined, for example:

image

If `$mod_args` is empty, it will just be an empty string, due to the leading and trailing ".

milachae:
@boegel 
I resolved all the suggestions

scripts/module_overview/README.md
```diff
@@ -0,0 +1,94 @@
+# Module overview
+
+A script to generate an table overview of all available module files inMarkDown format,
+which indicates on which clusters each module is available.
+
+## Requirements
+- Required Python packages are listed in the `requirements.txt`  file.
```
boegel, line 7:
```suggestion
- Required Python packages are listed in `requirements.txt` and `requirements_tests.txt`

File: scripts/module_overview/README.md [hunk 39]
```diff
@@ -0,0 +1,94 @@
...
+
+### Creating a virtual environment (optional)
+
+If the required Python packages are not available in your Python setup,
+you can easily create a dedicated virtual environment as follows:
+
+```shell
+python -m venv module_overview_venv
+source module_overview_venv/bin/activate
+pip install -r requirements.txt
```
boegel, line 19:
```suggestion
pip install -r requirements.txt
pip install -r requirements_tests.txt

File: scripts/module_overview/module_overview.py [hunk 40]
```diff
@@ -0,0 +1,223 @@
...
+
+def generate_general_overview() -> None:
+    """
+    Generate the general overview in a markdown file.
+    It generates a list of all the available software and indicates on which cluster it is available.
+    """
+    md_file = MdUtils(file_name='module_overview.md', title='Overview of available modules per cluster')
+    data = modules_ugent()
+    generate_module_table(data, md_file)
+    md_file.create_md_file()
```
boegel, line 218: also print path to file that was created?

boegel (changes requested):

boegel (changes requested):

PR #540, opened 2023-08-09
Fix python macros

### LINKED ISSUE #537: python macros like `{{ current_year }}` not working
See [example](https://docs.hpc.ugent.be/Linux/easybuild/?h=current+year#installing-supported-software)

xerbalind:
Closes #537

### DIFF
--- a/custom_plugin/custom_plugin.py
+++ b/custom_plugin/custom_plugin.py
@@ -42,7 +42,11 @@ def gen_content_from_macros():
             mod_name = f"{scripts}.{stem}"
             mod = importlib.import_module(mod_name)
             fun = getattr(mod, stem)
-            outputs[fn] = fun()
+
+            # set global variable name used in docs as the filename without the .py extension
+            variable_name = path.splitext(fn)[0]
+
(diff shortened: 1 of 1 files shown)


custom_plugin/custom_plugin.py
```diff
@@ -42,7 +42,7 @@ def gen_content_from_macros():
             mod_name = f"{scripts}.{stem}"
             mod = importlib.import_module(mod_name)
             fun = getattr(mod, stem)
-            outputs[fn] = fun()
+            outputs[fn[:-3]] = fun()
```
boegel, line 45: Hmm, can you explain why this fixes the bug exactly?

Why `3` and not `2` or `5`?

Seems like we're hardcoding something here that may change and will result in breaking the macros again...

boegel: I that case I would make this more explicit, by i) adding a comment to explain why it's needed, ii) using `os.path.splitext` instead:

```python
>>> os.path.splitext('foo.py')
('foo', '.py')
>>> os.path.splitext('foo.py')[0]
'foo'

Issue #543, opened 2023-08-10
Instant loading  is broken


xerbalind:
While exploring the website,  each time you navigate through the side menu or table of contents, the page undergoes a complete reload.

Within the code,  two types of markdown links exist:

```md
[link](../link)
```
OR

```md
[link](link.md)
```

The latter choice is significantly **preferred**.

When clicking a link from either option, the page loads instantly as anticipated. However, only after clicking on the the second type of link, from now on, when you navigate via the menu on the left, the page loads instantly (better). With the first link type the complete reloading issue persists.

I think it's something to do with os pick redirection

boegel:
Re-opening to get some feedback from @stdweird, but as far as we can tell removing the code in #547 doesn't cause trouble

xerbalind:
> Re-opening to get some feedback from @stdweird, but as far as we can tell removing the code in #547 doesn't cause trouble

It seems removing the code caused this problem: #558

PR #544, opened 2023-08-10
Add dark mode

xerbalind:
There is 'sometimes' still a flickering issue see #543


boegel:
This will allow me to make one of my favorite jokes during the HPC intro trainings ("light attracts bugs"), so great!

Looks pretty OK to me in dark mode on my system, so :+1:

PR #546, opened 2023-08-14
add section about using multiple nodes/cores

### LINKED ISSUE #541: add section in troubleshooting "Why does my job not run faster when using more cores and/or nodes?"
The difference in usage between Nodes/Cores is not really clear in the documentation.

- software does not run magically faster when more cores are available, the software must be able to actually use multiple cores (and may need to be told to use N>1 cores), if it supports this at all;

- for multi-node jobs: using resources of the other N-1 nodes requires using something like MPI (https://mpitutorial.com/tutorials), which usually involves starting the software via a command like `mpirun` or `mympirun` (see https://docs.hpc.ugent.be/mympirun), discovering which nodes are available (and how many cores per node), etc.

milachae:
Closes #541

### DIFF
--- a/mkdocs/docs/HPC/troubleshooting.md
+++ b/mkdocs/docs/HPC/troubleshooting.md
@@ -1,34 +1,55 @@
 # Troubleshooting
 
-## Why doesn't more ... lead to a faster job execution?
+## Why doesn't my job run faster when using more nodes and/or cores? { #job_does_not_run_faster }
+Requesting more resources for your job, more specifically using multiple cores and/or nodes, does not automatically imply that your job will run faster. There are various factors that determine to what extent these extra resources can be used and how efficiently they can be used. More information on this in the subsections below.
 
-### Cores
-When you want to make use of multiple cores, you also need to adapt your code to work with this. A normal, standard program will always run sequentially and you need to adapt it to really make use of the extra cores. You will always need to write some code to create, manage or synchronize multiple threads/processes. More on how to implement parallelization for you exact programming language can be found online.
+### Using multiple cores
+When you want to make use of multiple cores, you also need to adapt your code to work with this or use software that is capable of using multiple cores.
+Unless particular parallel programming paradigms like [OpenMP](https://www.openmp.org/about/openmp-faq/#WhatIs) threading (shared memory) or [MPI](https://en.wikipedia.org/wiki/Message_Passing_Interface) (distributed memory) are used, a program will run sequentially (on a single core).
+To use multiple cores, the software needs to be able to create, manage, and synchronize multiple threads or processes. More on how to implement parallelization for you exact programming language can be found online.
+Note that when using software that only uses *threads* to use multiple cores, there is no point in asking for multiple *nodes*, since with a multi-threading (shared memory) approach you can only use the resources of a single node.
 
-It could also be possible that more cores are simply not needed. You can test this by increasing the amount of cores and look at the speedup this increase gave you. For example, test with 2, 4, 16, 1/4 of all and all cores.
+Even if your software is able to use multiple cores, maybe there's no point in going beyond a single core, for example because the workload you are running is too small or doesn't parallelize well.
+You can test this by increasing the amount of cores step-wise, and look at the speedup you gain. For example, test with 2, 4, 16, a quarter, half and all available cores.
 
-Some **other reasons** why more cores could lead to no increase could be:
+Other reasons why using more cores may not lead to a (significant) speedup include:
 
-- **Overhead:** When you use multithreading or multiprocessing, you can't expect that double the amount of cores will double the performance. This is due to the fact that it also needs time to create, manage and synchronize the threads/processes. When the overhead exceeds the speedup you receive by parallelization, it will not benefit the performance. For example, this can happen when you split you program in to many pieces to run in parallel. Creating a thread/process for this task will take longer than actually running the task itself.
+- **Overhead:** When you use multi-threading (OpenMP) or multi-processing (MPI), you should *not* expect that doubling the amount of cores will result in a 2x speedup. This is due to the fact that time is needed to create, manage and synchronize the threads/processes.
+When this "bookkeeping" overhead exceeds the time gained by parallelization, you will not observe any speedup (or even see slower runs).
+For example, this can happen when you split you program in too many (tiny) tasks to run in parallel - creating a thread/process for each task will take longer than actually running the task itself.
 
-- **Amdahl's Law:** Amdahl's law is often used in parallel computing to predict the theoretical speedup when using multiple processors. The law states that "the overall performance improvement gained by optimizing a single part of a system is limited by the fraction of time that the improved part is actually used". For example, if a program needs 20 hours to complete using a single thread, but a one-hour portion of the program cannot be parallelized, therefore only the remaining 19 hours execution time can be parallelized. Regardless of how many threads are devoted to a parallelized execution of this program, the minimum execution time is always more than 1 hour. So when you reach this theoretical limit, more cores will not solve your problem.
+- **[Amdahl's Law](https://en.wikipedia.org/wiki/Amdahl%27s_law)** is often used in parallel computing to predict the maximum achievable (theoretical) speedup when using multiple cores. It states that "*the overall performance improvement gained by optimizing a single part of a system is limited by the fraction of time that the improved part is actually used*". For example, if a program needs 20 hours to complete using a single core, but a one-hour portion of the program can not be parallelized, only the remaining 19 hours of execution time can be sped up using parallelization. Regardless of how many cores are devoted to a parallelized execution of this program, the minimum execution time is always more than 1 hour. So when you reach this theoretical limit, using more cores will not help at all to speed up the computational workload.
 
-- **Resource contention:** When 2 or more threads/processes want to acces the same resources, they need to wait on each other. This is what we call resource contention. This results in that 1 thread/process will need to wait until the other one is is finished with this resource. When for example, each thread uses the same resource, it will definitely run slower than if it doesn't need to wait for the other threads to finish.
+- **Resource contention:** When two or more threads/processes want to access the same resource, they need to wait on each other - this is called resource contention. 
+As a result, 1 thread/process will need to wait until the other one is is finished using that resource. 
+When each thread uses the same resource, it will definitely run slower than if it doesn't need to wait for other threads to finish.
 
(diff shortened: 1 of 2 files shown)


mkdocs/docs/HPC/troubleshooting.md
```diff
@@ -1,6 +1,44 @@
...
+
+- **Software Limitations:** It could also be that your software is just not really optimized for parallelization. A perfect example of a software that is not really optimized for multithreading is python. Although this has improved over the years. This is due to the fact that threads are implemented in a way that multiple threads can't run at the same time. This is due to the global interpreter lock (GIL). Instead of using multithreading in python to speedup a CPU bound program, you should use multiprocessing instead. This uses processes instead of actual threads. These can speedup you're CPU bound programs a lot more IN PYTHON than threads can do, although they are much les efficiënt to create. In other programming languages, you would probably wan't to use threads.
```
boegel, line 18:
```suggestion
- **Software Limitations:** It is possible that the software you are using is just not really optimized for parallelization. A perfect example of a software that is not really optimized for multi-threading is Python. (although this has improved over the years).
  This is due to the fact that in Python threads are implemented in a way that multiple threads can't run at the same time, due to the [global interpreter lock (GIL)](https://en.wikipedia.org/wiki/Global_interpreter_lock). 
  Instead of using multi-threading in Python to speedup a CPU bound program, you should use multi-processing instead, 
  which uses multiple processes (multiple instances of the same program) instead of multiple threads in a single program instance.
  Multiprocessing can speed up you're CPU bound programs a lot more *in Python* than threads can do,
  even though they are much less efficient to create.
  In other programming languages (which don't have a GIL), you would probably still want to use threads, however.

File: mkdocs/docs/HPC/troubleshooting.md [hunk 2]
```diff
@@ -1,6 +1,44 @@
...
+
+- **no memory:** When memory bandwidth can't keep up with the increased core count, you will also see almost no improvements.
```
boegel, line 20:
```suggestion
- **Lack of sufficient memory**: When there not enough memory available, or not enough memory bandwidth, it is likely that you won't see a significant speedup when using more cores (since each thread or process most likely requires additional memory).
```

mkdocs/docs/HPC/troubleshooting.md
```diff
@@ -1,6 +1,44 @@
...
+
+More info about how you can run these on our HPC infrastructure can be found [here](multi_core_jobs.md).
+
+### Nodes
+When attempting to use multiple nodes in your parallelized program to enhance performance, It is a possibility that you notice no performance increase.
```
boegel, line 25:
```suggestion
When trying to use multiple (worker)nodes to improve the performance of your workloads,
you may not see (significant) speedup.

**Review by boegel — CHANGES_REQUESTED** [review]
Add one more small subsection, something like:

```
### How do I know if my software can run in parallel?

If you are not sure if the software you are using can efficiently use multiple cores or run across multiple nodes, you should check its documentation for instructions on how to run in parallel, or check for options that control how many threads/cores/nodes can be used.

If you can not find any information along those lines, the software you are using can probably only use a single core, and thus requesting multiple cores and/or nodes will only result in wasted sources.

mkdocs/docs/HPC/troubleshooting.md
```diff
@@ -1,6 +1,67 @@
+# Troubleshooting
 
+## Why doesn't my job run faster when using more nodes and/or cores? { #job_does_not_run_faster }
```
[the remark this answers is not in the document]
milachae:
```suggestion
## Why does my job not run faster when using more nodes and/or cores? { #job_does_not_run_faster }
```

milachae:
@boegel 
I applied all changes.

mkdocs/docs/HPC/troubleshooting.md
```diff
@@ -1,6 +1,72 @@
...
+
+### How do I know if my software can run in parallel?
+If you are not sure if the software you are using can efficiently use multiple cores or run across multiple nodes, you should check its documentation for instructions on how to run in parallel, or check for options that control how many threads/cores/nodes can be used.
+
+If you can not find any information along those lines, the software you are using can probably only use a single core and thus requesting multiple cores and/or nodes will only result in wasted sources.
+
+### Using multiple cores
+When you want to speed up your jobs by requesting multiple cores, you also need to use software that is actually capable of using them (and use them efficiently, ideally).
+Unless particular parallel programming paradigms like [OpenMP](https://www.openmp.org/about/openmp-faq/#WhatIs) threading (shared memory) or [MPI](https://en.wikipedia.org/wiki/Message_Passing_Interface) (distributed memory) are used, a program will run sequentially (on a single core).
```
boegel, line 13:
```suggestion
Unless particular a parallel programming paradigm like [OpenMP](https://www.openmp.org/about/openmp-faq/#WhatIs) threading (shared memory) or [MPI](https://en.wikipedia.org/wiki/Message_Passing_Interface) (distributed memory) is used, software will run sequentially (on a single core).

```

(adding empty line in between so next part starts new paragraph in rendered docs)

mkdocs/docs/HPC/troubleshooting.md
```diff
@@ -1,6 +1,72 @@
...
+To use multiple cores, the software needs to be able to create, manage, and synchronize multiple threads or processes. More on how to implement parallelization for you exact programming language can be found online.
+Note that when using software that only uses *threads* to use multiple cores, there is no point in asking for multiple *nodes*, since with a multi-threading (shared memory) approach you can only use the resources of a single node.
```
boegel, line 15:
```suggestion
Note that when using software that only uses *threads* to use multiple cores, there is no point in asking for multiple *nodes*, since with a multi-threading (shared memory) approach you can only use the resources (cores, memory) of a *single* node.

File: mkdocs/docs/HPC/troubleshooting.md [hunk 7]
```diff
@@ -1,6 +1,72 @@
...
+
+Even if your software is able to use multiple cores, maybe there is no point in going beyond a single core, for example because the workload you are running is too small or does not parallelize well.
```
boegel, line 17:
```suggestion
Even if your software is able to use multiple cores, maybe there is no point in going beyond a single core or a handful of cores, for example because the workload you are running is too small or does not parallelize well.

File: mkdocs/docs/HPC/troubleshooting.md [hunk 8]
```diff
@@ -1,6 +1,72 @@
...
+
+- **Resource contention:** When two or more threads/processes want to access the same resource, they need to wait on each other - this is called resource contention. 
+As a result, 1 thread/process will need to wait until the other one is is finished using that resource. 
+When each thread uses the same resource, it will definitely run slower than if it doesn't need to wait for other threads to finish.
+
+- **Software Limitations:** It is possible that the software you are using is just not really optimized for parallelization. An example of a software that is not really optimized for multi-threading is Python (although this has improved over the years).
```
boegel, line 32:
```suggestion
- **Software limitations:** It is possible that the software you are using is just not really optimized for parallelization. An example of a software that is not really optimized for multi-threading is Python (although this has improved over the years).

File: mkdocs/docs/HPC/troubleshooting.md [hunk 9]
```diff
@@ -1,6 +1,72 @@
...
+This is due to the fact that in Python threads are implemented in a way that multiple threads can not run at the same time, due to the [global interpreter lock (GIL)](https://en.wikipedia.org/wiki/Global_interpreter_lock). 
+Instead of using multi-threading in Python to speedup a CPU bound program, you should use multi-processing instead,
```
boegel, line 34:
```suggestion
Instead of using multi-threading in Python to speed up a CPU bound program, you should use multi-processing instead,

File: mkdocs/docs/HPC/troubleshooting.md [hunk 10]
```diff
@@ -1,6 +1,72 @@
...
+As a result, 1 thread/process will need to wait until the other one is is finished using that resource. 
+When each thread uses the same resource, it will definitely run slower than if it doesn't need to wait for other threads to finish.
+
+- **Software Limitations:** It is possible that the software you are using is just not really optimized for parallelization. An example of a software that is not really optimized for multi-threading is Python (although this has improved over the years).
+This is due to the fact that in Python threads are implemented in a way that multiple threads can not run at the same time, due to the [global interpreter lock (GIL)](https://en.wikipedia.org/wiki/Global_interpreter_lock). 
+Instead of using multi-threading in Python to speedup a CPU bound program, you should use multi-processing instead, 
+which uses multiple processes (multiple instances of the same program) instead of multiple threads in a single program instance.
+Multiprocessing can speed up your CPU bound programs a lot more *in Python* than threads can do, even though they are much less efficient to create.
```
boegel, line 36:
```suggestion
Using multiple processes can speed up your CPU bound programs a lot more *in Python* than threads can do, even though they are much less efficient to create.

File: mkdocs/docs/HPC/troubleshooting.md [hunk 11]
```diff
@@ -1,6 +1,72 @@
...
+In other programming languages (which don't have a GIL), you would probably still want to use threads.
+
+- **[Affinity and core pinning](https://www.admin-magazine.com/HPC/Articles/Processor-Affinity-for-OpenMP-and-MPI)**: 
+Even when the software you are using is able to efficiently use multiple cores, you may not see any speedup (or even a significant slowdown).
+This could be due to threads or processes that are not pinned to specific cores and keep hopping around between cores, or because the pinning is done incorrectly and several threads/processes are being pinned to the same cores, and thus keep "fighting" each other.
```
boegel, line 41:
```suggestion
This could be due to threads or processes that are not pinned to specific cores and keep hopping around between cores, or because the pinning is done incorrectly and several threads/processes are being pinned to the same core(s), and thus keep "fighting" each other.

File: mkdocs/docs/HPC/troubleshooting.md [hunk 12]
```diff
@@ -1,6 +1,72 @@
...
+
+- **Lack of sufficient memory**: When there not enough memory available, or not enough memory bandwidth, it is likely that you will not see a significant speedup when using more cores (since each thread or process most likely requires additional memory).
```
boegel, line 43:
```suggestion
- **Lack of sufficient memory**: When there is not enough memory available, or not enough memory bandwidth, it is likely that you will not see a significant speedup when using more cores (since each thread or process most likely requires additional memory).

boegel (changes requested):

mkdocs/docs/HPC/troubleshooting.md
milachae: https://en.wikipedia.org/wiki/Speedup

milachae:
@boegel
I applied all changes.

boegel (approved):

Issue #553, opened 2023-08-21
Overview modules


milachae:
Make an overview of all available modules on the different clusters.

- [ ] Automatically generate markdown table with all the available modules based on the lmod output
    - [ ] fetch data from lmod
    - [ ] convert to mk table
- [ ] Change to dynamic searchable datatable
    - [ ] put the data in a JSON file
    - [ ] Load data and populate table
- [ ] Add Overview for each software
    - [ ] Table with different versions on the different clusters (toolchains)

boegel:
Just so we have a record of this: we already have a JSON file that is automatically updated (along with the Lmod cache), which provides *most* information that is necessary to create a software overview, it can be obtained from `/apps/gent/lmodcache/modulemap.json` when logged in to the HPC-UGent login nodes.
Information missing there is the description + homepage that you get for individual modules via `module show` which would be nice to include in the software-specific overview of available versions.

boegel:
@stdweird Do you prefer only having a software overview available in the web portal as opposed to in the docs (which are public & hosted on GitHub)?
I mean outside of forms to create a desktop apps & such, so a more "static" table overview (but which is somewhat searchable, like the table-with-filter approach that @milachae has a proof-of-concept of), which eventually also includes extensions (like `numpy` which is included in the `SciPy-bundle` module, R packages, etc.).

Shipping all necessary data to produce an overview of installed modules isn't that huge, the page still loads quickly enough in practice (@milachae has been experimenting with that).

stdweird:
@boegel just pointing out some options. if it's fast enough and does what you want, go for it

PR #554, opened 2023-08-21
Overview of available software (with filter)

milachae:
#553


config/templates/hpc.template
```diff
@@ -6,6 +6,7 @@ docs_dir: docs/HPC
 
 nav:
   - Welcome: index.md
+  - Overview modules: module_overview.md
```
boegel, line 9: Rename:

```suggestion
  - Available software: available_software.md
```

I think the plan was to exclude this page from the menu for now?

mkdocs/docs/HPC/module_overview.md
```diff
@@ -0,0 +1,10 @@
+---
+hide:
+  - toc
+---
+
+# Module overview
```
boegel, line 6:
```suggestion
# Available software (via modules)

File: scripts/module_overview/module_overview.py [hunk 3]
```diff
@@ -220,6 +221,42 @@ def generate_general_overview() -> None:
...
+    json_data = {"clusters": list(modules.keys()), "modules": {}}
+    avail_software = get_unique_software_names(modules)
+    all_software = get_unique_software_names(np.concatenate(list(avail_software.values())))
+
+    # creates a list of booleans for each software that indicates
+    # if the software is available for the corresponding cluster.
+    for soft in all_software:
+        available = []
+        for cluster in json_data["clusters"]:
+            available.append(soft in avail_software[cluster])
```
boegel, line 240: Plan was to use `0` and `1` to make the JSON smaller?

```suggestion
            # use 0 and 1 rather than true/false to obtain smaller JSON file
            available.append(int(soft in avail_software[cluster]))

boegel (changes requested):
@milachae Can we also include a timestamp in the JSON, and include that in the module overview ("`Last updated at ...`")?

scripts/module_overview/module_overview.py
milachae: Yes, this is already implemented in #559

config/templates/hpc.template
milachae: Indeed, I will remove it

milachae:
> @milachae Can we also include a timestamp in the JSON, and include that in the module overview ("`Last updated at ...`")?

Indeed, I forgot about this. Will add it now.

boegel (approved):

PR #557, opened 2023-08-23
move linux tutorial

xerbalind:


config/others/linux-tutorial.yml
```diff
@@ -1,14 +1,7 @@
 site_name: Linux tutorial
-docs_dir: docs/linux-tutorial
+docs_dir: docs/linux
```
boegel, line 2: We should keep `linux-tutorial` as directory name here, to avoid breaking the existing URL https://hpcugent.github.io/vsc_user_docs/linux-tutorial/ ?

config/others/linux-tutorial.yml
```diff
@@ -1,14 +1,7 @@
...
 nav:
-  - Introduction: index.md
-  - Getting Started: getting_started.md
-  - Navigating: navigating.md
-  - Manipulating files and directories: manipulating_files_and_directories.md
-  - Uploading/downloading/editing files: uploading_files.md
-  - Beyond the basics: beyond_the_basics.md
-  - Common Pitfalls: common_pitfalls.md
-  - More on the HPC infrastructure: hpc_infrastructure.md
+  - Welcome: index.md
```
boegel, line 4: Wouldn't it be better to hide the navigation part entirely on that page, since all it does is provide a link to the new page?

mkdocs/docs/linux/index.md
```diff
@@ -0,0 +1,2 @@
+
+**The tutorial has been moved to [here](https://docs.hpc.ugent.be/linux-tutorial)**
```
boegel, line 2: I would expose the URL here rather than hiding it behind `here`, And let's add a page title too:

```md
# This tutorial has moved!

See <https://docs.hpc.ugent.be/linux-tutorial> .

Please update your bookmarks!

**Review by boegel — CHANGES_REQUESTED** [review]

File: mkdocs/docs/linux-tutorial/index.md [hunk 4]
```diff
@@ -0,0 +1,5 @@
+# This tutorial has moved!
```
boegel, line 1: Let's hide the nav bar on the left (and toc on the right) completely on this page:

```suggestion
---
hide:
  - navigation
  - toc
---

# This tutorial has moved!

boegel (changes requested):

PR #564, opened 2023-09-01
Module list eessi

milachae:
@boegel

You can run the script in EESSI mode by running following command:
`python available_software.py --eessi`

But it looks that there is still a problem, because it finds a lot of modules.
image

I suspect that it can't do the `module unuse $MODULEPATH` properly, but I was not yet able to solve that bug and don't have the time to look deeper into it.

I find all the correct folder by executing following 2 commands:
image

First, I tried to just run 2 ls commands, but they didn't work properly when executed from python:

```
ls -1d /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/*/!(intel|amd)
ls -1d /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/*/{amd,intel}/*
```


boegel:
@milachae I fixed the `module unuse $MODULEPATH` issue in 4168365 .

Some improvements can definitely be made for EESSI (like the way too long column names), but that can be dealt with in a follow-up PR, so merging this. Thanks a lot!

PR #569, opened 2023-09-12
remove cloud from deploy workflow

boegel:
this was missing from #568


alvarosimon:
@boegel ups I forgot about this

PR #570, opened 2023-09-19
kickstart docs for custom web portal apps

boegel:


mkdocs/docs/HPC/web_portal.md
```diff
@@ -346,3 +346,7 @@ menu item:
 
 Of course, this only affects your own web portal session (not those of
 others).
+
+## Custom apps
```
stdweird, line 350: ugent only?

boegel: The whole web portal page is UGent-only, but we can deal with that in a separate PR if we really want to.

The custom apps are even specific to certain user groups probably, we could clearly mention that.

This PR is just to get @epauwels going with adding docs for specific custom apps

PR #592, opened 2023-12-21
HPC tutorial: update python scripts

laraPPr:
I started updating the example scripts from the HPC tutorial because these are deprecated. The only place I could find them was in the latex branch. But maybe the example scripts should be managed somewhere else.


boegel:
@laraPPr Let's also copy these to an `examples` directory in `main` branch?

PR #599, opened 2024-03-11
Update HPC-UGent policies for externals

epauwels:


itkovian:
So basically push every industry user to Tier1, right?

epauwels:
> So basically push every industry user to Tier1, right?

Correct. And also remove the 1500 euro/year stimulus for other research institutes. They should also go to Tier1.

PR #610, opened 2024-07-02
added 'login node' after 'the'

lbarraga:
There are two instances where the word 'the' is not followed by a word. I think the missing words should be 'login node'

### DIFF
--- a/mkdocs/docs/HPC/connecting.md
+++ b/mkdocs/docs/HPC/connecting.md
@@ -433,49 +433,49 @@ Connection to {{ loginnode }} closed.
     LC_ALL=
     ```
 
-A **locale** is a set of parameters that defines the user's language, country and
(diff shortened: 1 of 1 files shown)


mkdocs/docs/HPC/connecting.md
```diff
@@ -474,7 +474,7 @@ again on your local machine:
 </code></pre>
 
 You can now log out, open a new terminal/shell on your local machine and
-reconnect to the , and you should not get these warnings anymore.
+reconnect to the login node, and you should not get these warnings anymore.
```
boegel, line 477: The indent is off here as well, this paragraph (and the ones above) all belong in the "`tip: Setting your Language right`", so fix that too while you're at it?

boegel (changes requested):

boegel (approved):

PR #614, opened 2024-07-03
copy example scripts used in docs from latex branch

boegel:
These scripts are used in the documentation,
and can be copied from `/apps/gent/tutorials/Intro-HPC/examples`


laraPPr (approved):
lgtm

PR #619, opened 2024-07-03
Du hidden files command

lbarraga:
This pull request suggests changing the `-s` and `*` for `--max-depth 1` in the last two examples on `du` usage. There are two reasons for this suggestion:

Firstly, the docs say:
 > If you want to see the size of ***any*** file or top-level subdirectory in the current directory, you could use the following command: `du -s -h *`

Because `*` does not expand hidden files and directories it could be misleading for users trying to figure out the space their files are taking up. In contrast, `--max-depth 1` does take this into account.  

Secondly, I overheard @hajgato say users who hit quota frequently overlook hidden files and directories when cleaning up disk space. Some unnecessary tickets could be avoided with the new commands


boegel:
Makes sense to me, what do you think @hajgato ?

PR #630, opened 2024-07-04
added note explaining the difference between storage and memory

lbarraga:
The difference between storage and memory might not be trivial for all users. I added a note explaining that, while also linking to the section on storage quota issues:

image


boegel (approved):
:+1:

PR #633, opened 2024-07-05
compile eat_cpu before running

lbarraga:
As the `eat_cpu` binary is not present in `examples/Fine-tuning-Job-Specifications`, it must first be compiled. I added this to the docs.

image

### DIFF
--- a/mkdocs/docs/HPC/fine_tuning_job_specifications.md
+++ b/mkdocs/docs/HPC/fine_tuning_job_specifications.md
@@ -515,9 +515,10 @@ The **uptime** command will show us the average load
 10:14:05 up 86 days, 12:01, 11 users, load average: 0.60, 0.41, 0.41
 </code></pre>
 
-Now, start a few instances of the "*eat_cpu*" program in the background,
+Now, compile and start a few instances of the "*eat_cpu*" program in the background,
 and check the effect on the load again:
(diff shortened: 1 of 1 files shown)


boegel (approved):

mkdocs/docs/HPC/fine_tuning_job_specifications.md
```diff
@@ -515,9 +515,10 @@ The **uptime** command will show us the average load
 10:14:05 up 86 days, 12:01, 11 users, load average: 0.60, 0.41, 0.41
 </code></pre>
 
-Now, start a few instances of the "*eat_cpu*" program in the background,
+Now, compile and start a few instances of the "*eat_cpu*" program in the background,
 and check the effect on the load again:
-<pre><code>$ <b>./eat_cpu&</b>
+<pre><code>$ <b>gcc eat_cpu.c -o eat_cpu</b>
```
boegel, line 520: By default `gcc` uses `-O0`, i.e. no optimizations at all.

That's a bad default imho, and we should try and make people aware of that when we have the chance, so let's change this to:

```
<pre><code>$ <b>gcc -O2 eat_cpu.c -o eat_cpu</b>
```

boegel (changes requested):

boegel (approved):

PR #658, opened 2024-07-30
Update modules in examples

laraPPr:
A number of the examples will fail because of the outdated modules that are no longer available on the cluster or loading intel toolchains on AMD system.

Linked to this OTRS ticket: [#202407236000017](https://otrsdict.ugent.be/otrs/index.pl?Action=AgentTicketZoom&TicketID=162880)

### DIFF
--- a/intro-HPC/examples/OpenFOAM/OpenFOAM_damBreak.sh
+++ b/intro-HPC/examples/OpenFOAM/OpenFOAM_damBreak.sh
@@ -2,7 +2,7 @@
 #PBS -l walltime=1:0:0
 #PBS -l nodes=1:ppn=4
 # check for more recent OpenFOAM modules with 'module avail OpenFOAM'
-module load OpenFOAM/6-intel-2018a
+module load OpenFOAM/11-foss-2023a
 source $FOAM_BASH
 # purposely not specifying a particular version to use most recent mympirun
 module load vsc-mympirun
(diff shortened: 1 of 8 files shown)


intro-HPC/examples/OpenFOAM/OpenFOAM_damBreak.sh
laraPPr, line 1: This will probably now fail. You can find a compatible example script for OpenFOAM 11 here https://github.com/EESSI/eessi-demo/tree/main/OpenFOAM

laraPPr:
@boegel Maybe the jobstudents can put some time into testing these scripts?

laraPPr:
Should I also remove the PQS and nwChem examples since these are not installed on the system?

intro-HPC/examples/OpenFOAM/OpenFOAM_damBreak.sh
EwDa291: As far as I can tell, the example seems to work if line 18 also changes to the following:

cp -r $FOAM_TUTORIALS/incompressibleVoF/damBreakLaminar/damBreak $WORKDIR

instead of 

cp -r $FOAM_TUTORIALS/multiphase/interFoam/laminar/damBreak/damBreak $WORKDIR

However this should be taken with a grain of salt as I am not familiar with the workings of openFOAM.
Making this change will result in results being saved and no errors being thrown.

laraPPr:
@boegel ready to merge now

boegel (approved):

PR #661, opened 2024-08-05
New FAQ for du command

EwDa291:


mkdocs/docs/HPC/FAQ.md
```diff
@@ -373,6 +373,19 @@ See also: [Your UGent home drive and shares](running_jobs_with_input_output_data
 {% endif %}
 
 
+### My home directory is (almost) full, and I don't know why
+
+Your home directory might be full without looking like it due to hidden files. Hidden files and subdirectories have a name starting with a dot and do not show up when using a normal `ls`. If you want to check where the storage in your home directory is used, you can make use of the [`du` command](https://docs.hpc.ugent.be/Linux/running_jobs_with_input_output_data/#check-your-quota) to find out what the largest files and subdirectories are:
```
boegel, line 378: Small tweaks to text + put each sentence on a new line for easier editing:

```suggestion
Your home directory might be full without looking like it due to hidden files.
Hidden files and subdirectories have a name starting with a dot and do not show up when running `ls`.
If you want to check where the storage in your home directory is used, you can make use of the [`du` command](https://docs.hpc.ugent.be/Linux/running_jobs_with_input_output_data/#check-your-quota) to find out what the largest files and subdirectories are:

File: mkdocs/docs/HPC/FAQ.md [hunk 2]
```diff
@@ -373,6 +373,19 @@ See also: [Your UGent home drive and shares](running_jobs_with_input_output_data
...
+
+```shell
+du -h --max-depth 1 $VSC_HOME | egrep '[0-9]{3}M|[0-9]G'
+```
+
+The `du` command returns the size of every file and subdirectory in the $VSC_HOME directory. This output is then piped into an [`egrep`](https://docs.hpc.ugent.be/Linux/linux-tutorial/beyond_the_basics/?h=grep#searching-file-contents-grep) to filter the lines to the ones that matter the most.
+
+The `egrep` command will only let entries that match with the specified regular expression `[0-9]{3}M|[0-9]G` through, which corresponds with files that consume between 100 and 1024 MB or more than 1 GB.
```
boegel, line 386: It's actually 999MB, since it's 3 digits :)

boegel (changes requested):

mkdocs/docs/HPC/FAQ.md
EwDa291: I thought so too, but I used a tool to check which strings matched with the regex and 1000M also matched, the first digit just wasn't part of the matching part. The same thing worked for files larger than 10 GB

boegel: Yeah, right OK.

To avoid confusion, maybe just reword then to "files that consume 100 MB or more"?

EwDa291: Will do ;)

boegel (approved):

PR #670, opened 2024-08-23
add and document `vsc-venv` tool for creating Python virtual environments

lbarraga:
`vsc-venv` is a script that encapsulates the creation and management of Python virtual environments. 
This avoids multiple issues with the default `venv` included in Python (`python -m venv`).

One issue is that a virtual environment created for one cluster might not work on another. 
Additionally, when activating a virtual environment, 
the same centrally installed modules used during its creation must also be loaded.
the `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way, 
while guaranteeing the same module environment.


boegel:
@lbarraga As discussed, please add a small README file with some basic usage info, and perhaps a concrete example, so we (@laraPPr, @hajgato, and myself) can try this out ourselves hands-on

lbarraga:
## TODO

- [ ] When activating, modules need to be loaded. 
- [ ] Warning if modules are active when creating env (if they are not in the modules.sh?)
- [ ] updating existing virtual environments

lbarraga:
After thinking for a while, copying the python venv commands (create, activate, install) does not seem like the best course of action. The more commands, the more the user can mess up and the more edge cases.

Users just want to provide `modules.sh` and `requirements.txt` and start working. I propose only *one* command to enter the environment. This command just makes sure only the modules in `modules.sh` are loaded and `requirements.txt` is installed.

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo "  install <requirements>   Install packages from the specified requirements file."
+  echo "                           The virtual environment must be activated before running this command."
+  echo "  activate                 Activate the virtual environment. You need to source this command."
+  echo ""
+  echo "Example Usage:"
+  echo "  $ $0 create my_modules.sh"
+  echo "  $ source $0 activate"
+  echo "  $ $0 install requirements.txt"
+  echo "  $ python my_script.py"
+  echo "  $ deactivate"
```
stdweird, line 21: for consistency, i would make this `$0 deactivate`. and you never know you want to gather some more metadata on deactivation, eg the loaded modules on deactivation.

boegel:
> After thinking for a while, copying the python venv commands (create, activate, install) does not seem like the best course of action. The more commands, the more the user can mess up and the more edge cases.
> 
> Users just want to provide `modules.sh` and `requirements.txt` and start working. I propose only _one_ command to enter the environment. This command just makes sure only the modules in `modules.sh` are loaded and `requirements.txt` is installed.

I agree, keep the usage as basic as possible, and just auto-detect whether create/install still needs to happen or not (and do so, if necessary).

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+To deactivate the virtual environment, run:
+
+```bash
+source ./vsc-venv.sh deactivate
+```
+
+## Example
+
+Suppose we are on the login node, and we want to make a virtual environment for doduo, which is loaded by default.
+The following files are present in the current directory:
```
stdweird, line 37: mandatory?

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+**modules.sh:**
+```bash
+ml Python/3.12.3-GCCcore-13.3.0
```
stdweird, line 41: this is any bash code?

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+and **requirements.txt:**
```
stdweird, line 44: add a ref to the supported syntax (eg version locks/limits)

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+requests
+```
+
+We run the following commands to enter the environment
+
+```bash
+source ./vsc-venv.sh activate requirements.txt modules.sh
+```
+
+As this creates the virtual environment for the first time, the `venvs` folder is created.
```
stdweird, line 55: created in the current dir?

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+We run the following commands to enter the environment
+
+```bash
+source ./vsc-venv.sh activate requirements.txt modules.sh
+```
+
+As this creates the virtual environment for the first time, the `venvs` folder is created. 
+Within this folder, a new folder is created for the virtual environment `venv-RHEL8-zen2`.
```
stdweird, line 56: it's a first step, but doesn't protect against incompatible toolchains

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+### joltik
+
+If we want to create a virtual environment for joltik, we can run the following commands:
+
+```bash
+module swap cluster/joltik
+qsub -I
+source ./vsc-venv.sh activate requirements.txt modules.sh
```
stdweird, line 73: maybe add the `cd ...` to set the proper path?

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
...
+
+echo_info() { echo -e "\e[32m[INFO] $1\e[0m"; }
+echo_warning() { echo -e "\e[33m[WARNING] $1\e[0m"; }
+echo_error() { echo -e "\e[31m[ERROR] $1\e[0m"; }
+
+
+# ============================ Main functions ============================
+
+activate() {
+  REQUIREMENTS_FILE="$1"
```
stdweird, line 27: lowercase variables, and use `local ...` to make them scoped in the function
```bash
local requiremets_file  modules_script
requirements_file="$1"
...
```

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
...
+  MODULES_SCRIPT="$2"
+
+  VENV_LOCATION=$(realpath -m "venvs/venv-${VSC_OS_LOCAL}-${VSC_ARCH_LOCAL}") # full path of venv
+
+  # === Step 0: Warn user if they have modules loaded === #
+
+  N_LOADED_MODULES=$(echo "$LOADEDMODULES" | tr ':' '\n' | wc -l)
+  if [ "$N_LOADED_MODULES" -gt 4 ]; then # 4 is the number of modules loaded by default
```
stdweird, line 35: there is no other way to test this?

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
...
+    echo_warning "You have loaded modules in the current shell. If you want to use these modules, please provide a modules script as the second argument."
```
stdweird, line 36: shoudl clarify that these are modules other then the cluster modules

scripts/vsc-venv-wrapper/vsc-venv.sh
boegel: I wouldn't count how many modules are loaded, I would filter out the `cluster/` and `env/*` modules, and then complain if the list is non-empty.

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
...
+  echo_info "Purging currently loaded modules."
+  module purge
+
+  # === Step 2: Load Modules if module script present === #
+
+  if [ -n "$MODULES_SCRIPT" ]; then # If module script not empty
+
+    echo_info "Loading modules from '$MODULES_SCRIPT'"
+
+    if ! source "$MODULES_SCRIPT"; then # If the module script could not be loaded
```
stdweird, line 51: how trustworthy is this? maybe wrap in `set +e ... set -e`?

scripts/vsc-venv-wrapper/README.md
boegel: If the exit code is checked for the `source modules.sh` call, then that's sufficient?

Anything beyond that it outside of scope for this helper script imho

scripts/vsc-venv-wrapper/vsc-venv.sh
boegel: I should also print the list of modules that are loaded but shouldn't be (so without `cluster/*` and `env/*`, those are fine)

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,82 @@
...
+## Usage
+
+A virtual environment can be activated by running the following command:
+
+```bash
+source ./vsc-venv.sh activate <requirements> [module_script]
+```
+
+Here, `requirements` is the path to a file containing the python dependencies to install in the virtual environment.
+An optional module script can be provided to load the required modules before creating the virtual environment.
```
stdweird, line 22: i don't get the optional part. if it is missing, the system python is being used, no?

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
...
+
+  else
+    echo_info "No module script provided. Proceeding without extra modules."
+  fi
+
+  # === Step 3: Create Virtual Environment if not yet present === #
+
+  echo_info "Creating virtual environment at $VENV_LOCATION"
+  # Will automatically make the venvs folder and venv, does nothing if they already exist
+  if ! python -m venv "$VENV_LOCATION"; then
```
stdweird, line 66: you should log the python that is being used: it's full path and the version

scripts/vsc-venv-wrapper/vsc-venv.sh
boegel: What do you mean?
Are you worried about a `modules.sh` script where the 2nd `module load` command fails, but the 3rd works fine?

In that case, yes, we need `set -e` so it fails on the first problem occurring, not only if the last command in `modules.sh` fails.

scripts/vsc-venv-wrapper/README.md
boegel: Hmm, maybe it's better to have a `modules.txt`, and then we load the modules one by one ourselves?

That also makes error checking easier.

scripts/vsc-venv-wrapper/README.md
boegel: Yes, and perhaps that should come with a warning.

There are use cases where using system Python is OK. EasyBuild for example, or ReFrame.

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,97 @@
+SCRIPT_NAME=${BASH_SOURCE[0]} # $0 cannot be used as it gives '-bash' when sourced
+
+usage() {
+  echo "Usage: source $SCRIPT_NAME {activate <requirements.txt> [modules.sh] | deactivate}"
```
stdweird, line 4: i think that this behaviour is not ok.
for a new venv, you need a module.sh to at least set the python version

but i would store the modules.sh and requirements.txt in the venv dir; and the when these are not provided; you reuse that one (well, at least the modules.sh) instead of doing nothing.
having the modules.sh and requiremts.tx you can later also implement a clone command

i also don't understand why the requirements.txt is mandtory: once initialised, to reuse the same environment, you don't need it anymore?

scripts/vsc-venv-wrapper/README.md
stdweird: if the venv does not exist, it should be mandatory and stored after venv creation

if the venv exists, it should be optional. if is not provided, reuse the one that was used during creation. if it is provided and different from original one, hmmm. a bit tricky; but you could overwrite the original one?

so for system python you can use an empty file or eg pass the 'SYSTEM' magic constant (but i would just do empty file)

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: I tried to make the script as fool-proof as possible. Copying the `modules.sh` and `requirements.txt` and making it optional could open the user up for unexpected behavior. The user would need to remember to initialize the script with the requirements every time packages are added. 

A solution could be to assume the `requirements.txt` and `modules.sh` are present in the root of the project (which is the case most of the time). That way, the user doesn't need to type out the requirements and modules every time.

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: Fixed in [lower case local variables](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/b228814481ce5cc861b3a4838562cc5b3fad6357)

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: Fixed in [loaded modules warning calculated by list + print of extra modules](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/5e0f1c6f31b0e995656449a0d684704b8d4b7010)

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: Fixed in [loaded modules warning calculated by list + print of extra modules](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/5e0f1c6f31b0e995656449a0d684704b8d4b7010)

scripts/vsc-venv-wrapper/README.md
lbarraga: Fixed in [warning when system python is used + logging of python version](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/3b924f433e971bde5ca1ebf458e55b9ecb08ccc2)

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: Log version in [warning when system python is used + logging of python version](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/3b924f433e971bde5ca1ebf458e55b9ecb08ccc2)

scripts/vsc-venv-wrapper/README.md
lbarraga: Added in [docs: add 'cd my_project'](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/bbc70d130f331b672fba6ad35911a81d439db9c5)

scripts/vsc-venv-wrapper/README.md
lbarraga: Clarified in [docs: venvs folder created in current directory](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/bd59341d41112ae206e375867532e2256a30b2ec)

scripts/vsc-venv-wrapper/README.md
lbarraga: Added in [requirements.txt docs referene added](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/210cb01434b465a6ec2b261294d44d896ea4e4b7)

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: This is indeed not trustworthy, but will be fixed with a `modules.txt` that is loaded line per line

lbarraga: Fixed with [5f05699](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/5f05699138ff64e6fe2e79e5cf107826c3a0665c). Modules are now loaded one by one. The script stops when a `module load` fails.

scripts/vsc-venv-wrapper/README.md
lbarraga: Done in [5f05699](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/5f05699138ff64e6fe2e79e5cf107826c3a0665c)

lbarraga:
In [warning when creating or activating on incompatible cluster](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/0136bafe30f57c4283513e18cb66f2a24cac190f), I added a warning when the user tries to activate or create a virtual environment on a host which is incompatible with the current loaded cluster. For example, swapping to joltik on the login node and activating a virtual environment is met with following output:

```bash
$ module swap cluster/joltik
$ source ./vsc-venv activate requirements.txt modules.txt
[WARNING] The OS or architecture of the current host (doduo) does not match that of the loaded cluster (joltik).
[WARNING] Creating or activating a virtual environment on doduo with modules optimized for joltik may cause issues.
[WARNING] to make a virtual environment on the joltik cluster, start an interactive session on that cluster:
[WARNING]   $ module swap cluster/joltik
[WARNING]   $ qsub -I
[WARNING]   $ cd /path/to/your/project
[WARNING] Then run the activate command again.
[WARNING] Exiting...
```

boegel:
A couple of small tweaks to make:

* make sure `--help` and `--version` options work; `--help` should just print a one-liner, and point to URL in docs (which can be created later);
* add detection whether script is sourced or not, it must be, so something like:
```shell
if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
    echo_error "This script must be sourced! Use: source $0"
fi
```

scripts/vsc-venv-wrapper/vsc-venv.sh
boegel:
```suggestion
  echo "Usage: source $SCRIPT_NAME {activate <requirements.txt> [modules.txt] | deactivate}"
```

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo ""
+  echo "Commands:"
+  echo "  activate <requirements_file> [modules_file] Activate the environment using the specified requirements file."
```
boegel, line 7:
```suggestion
  echo "  activate <requirements.txt> [modules.txt] Activate the environment using the specified requirements file."

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 16]
```diff
@@ -0,0 +1,182 @@
...
+  fi
+
+  for line in "${lines[@]}"; do # Loop over each line in the array
+
+    # Skip empty lines
+    if [[ -z "$line" ]]; then
+      continue
+    fi
+
+    if module load "$line"; then
```
boegel, line 39: Remove the quotes, so that it works fine if multiple module names are listed on 1 line in `modules.txt`:
```suggestion
    if module load $line; then

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 17]
```diff
@@ -0,0 +1,182 @@
...
+    echo_info "Modules loaded successfully"
+
+  else
+    echo_info "No module file provided. Proceeding without extra modules."
+  fi
+
+  # === Step 4: Create Virtual Environment if not yet present === #
+
+  if [ "$(which python)" = "/usr/bin/python" ]; then
+    echo_warning "System python used. Consider loading a specific python module through the modules file."
```
boegel, line 145:
```suggestion
    echo_warning "System python used. Consider loading a specific Python module through the modules file."

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 18]
```diff
@@ -0,0 +1,182 @@
...
+  fi
+
+  # === Step 5: Activate Virtual Environment === #
+
+  echo_info "Activating virtual environment"
+  source "$venv_location/bin/activate"
+
+  # === Step 6: Install Requirements === #
+
+  echo_info "Installing requirements from '$requirements_file'"
```
boegel, line 163: Check missing here to ensure that `${requirements_file}` really exists (or maybe do it higher up, as early as possible)

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo "Example Usage:"
+  echo "  $ source $SCRIPT_NAME activate requirements.txt modules.txt"
+  echo "  $ python my_script.py"
+  echo "  $ source $SCRIPT_NAME deactivate"
+  return 1
+}
+
+echo_info() { echo -e "\e[32m[INFO] $1\e[0m"; }
+echo_warning() { echo -e "\e[33m[WARNING] $1\e[0m"; }
+echo_error() { echo -e "\e[31m[ERROR] $1\e[0m"; }
```
boegel, line 21:
```suggestion
echo_info() { echo -e "\e[32m[vsc-venv: INFO] $1\e[0m"; }
echo_warning() { echo -e "\e[33m[vsc-venv: WARNING] $1\e[0m"; }
echo_error() { echo -e "\e[31m[vsc-venv: ERROR] $1\e[0m"; }
```

boegel:
Think about what to do if `activate` is done twice in a row, without `deactivate` in between. Error? Auto-run `deactivate` first?

boegel:
Print a warning when `requirements.txt` includes lines that don't specify a version (range)

boegel:
In the docs, mention `python -V`, `pip list`, `module list` as ways to check what's in the active environment.

boegel:
To consider: define alias/overrule `module` and `pip` to prevent that people manually do `module load` or `pip install` ?
Ideally in a way that `module list` and `pip list` still work fine :)

boegel (changes requested):

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+```bash
+source ./vsc-venv.sh --activate --requirements <requirements> [--modules module_file]
+```
+
+Here, `requirements` is the path to a file containing the python dependencies to install in the virtual environment.
+An optional `module_file` can be provided, which lists the modules to load before activating the virtual environment.
+
+Automatically, the modules are loaded and the environment is activated. 
+When running this command for the first time, the dependencies from the requirement file are installed.
```
boegel, line 25:
```suggestion
When running this command for the first time, the dependencies from the requirements file are installed.

File: scripts/vsc-venv-wrapper/README.md [hunk 21]
```diff
@@ -0,0 +1,95 @@
...
+
+Now, the software can be run and use packages installed in the virtual environment, along with centrally installed modules.
```
boegel, line 27:
```suggestion
Now, the software can be run and Python packages installed in the virtual environment can be used, along with software provided via centrally installed modules.

File: scripts/vsc-venv-wrapper/README.md [hunk 22]
```diff
@@ -0,0 +1,95 @@
...
+
+To deactivate the virtual environment, run:
+
+```bash
+source ./vsc-venv.sh --deactivate
+```
+
+## Example
+
+Suppose we are on the login node, and we want to make a virtual environment for doduo, which is loaded by default.
```
boegel, line 44:
```suggestion
Suppose we are on the HPC-UGent Tier-2 login nodes, and we want to make a virtual environment for doduo, the default cluster.
```

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,95 @@
...
+the same centrally installed modules used during its creation must also be loaded.
+the `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way, 
+while guaranteeing the same module environment.
+
+## Usage
+
+A virtual environment can be activated by running the following command:
+
+```bash
+source ./vsc-venv.sh --activate --requirements <requirements> [--modules module_file]
```
boegel, line 18: Change all occurrences of `./vsc-venv.sh` to `module load vsc-env`, and just using `source vsc-venv` (no `./`).

We'll install the `vsc-env` module with versioning.

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,95 @@
...
+For more info on the `requirements.txt` file, 
+see the [pip documentation](https://pip.pypa.io/en/stable/reference/requirements-file-format/).
+
+We run the following commands to enter the environment
+
+```bash
+source ./vsc-venv.sh --activate --requirements requirements.txt --modules modules.txt
+```
+
+As this creates the virtual environment for the first time, the `venvs` folder is created in the current directory.
```
boegel, line 67:
```suggestion
As this creates the virtual environment for the first time, a `venvs` subdirectory is created in the current directory.

File: scripts/vsc-venv-wrapper/README.md [hunk 25]
```diff
@@ -0,0 +1,95 @@
...
+see the [pip documentation](https://pip.pypa.io/en/stable/reference/requirements-file-format/).
+
+We run the following commands to enter the environment
+
+```bash
+source ./vsc-venv.sh --activate --requirements requirements.txt --modules modules.txt
+```
+
+As this creates the virtual environment for the first time, the `venvs` folder is created in the current directory. 
+Within this folder, a new folder is created for the virtual environment `venv-RHEL8-zen2`.
```
boegel, line 68:
```suggestion
Within `venvs/`, an additional subdirectory is created for the virtual environment: `venv-RHEL8-zen2`.
```

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Now, python 3.12 is loaded and the user can use the `requests` package, along with the packages from the SciPy bundle.
```
boegel, line 70: Unclear where `requests` comes from, just use `numpy` provided by `SciPy-bundle` module as example:
```suggestion
Now, Python 3.12 is loaded and both the `numpy` (provided by the `SciPy-bundle` module) and `requests` Python packages can be used.

File: scripts/vsc-venv-wrapper/README.md [hunk 27]
```diff
@@ -0,0 +1,95 @@
...
+
+To deactivate the virtual environment, run:
+
+```bash
+source ./vsc-venv.sh --deactivate
+```
+
+### joltik
+
+If we want to create a virtual environment for joltik, we can run the following commands:
```
boegel, line 80: Don't use `joltik` as example (it's old, and a GPU cluster), use `donphan` instead:
```suggestion
If we want to create a virtual environment for another cluster, say donphan, we can use `vsc-venv` in an interactive job:
```

scripts/vsc-venv-wrapper/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+```bash
+module swap cluster/joltik
```
boegel, line 83:
```suggestion
module swap cluster/donphan

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 29]
```diff
@@ -0,0 +1,252 @@
...
+}
+
+# ============================ Main ============================
+
+SCRIPT_NAME=${BASH_SOURCE[0]} # $0 cannot be used as it gives '-bash' when sourced
+
+
+# This script must be sourced
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    echo_error "This script must be sourced! Check the usage with 'source $SCRIPT_NAME --help'"
```
boegel, line 232:
```suggestion
    echo_error "This script must be sourced! Check the usage with 'source $(basename $SCRIPT_NAME) --help'"
```

This produces a better error message:
```
/path/to/vsc-venv: [ERROR] This script must be sourced! Check the usage with 'source vsc-venv --help'
```

instead of

```
/path/to/vsc-venv: [ERROR] This script must be sourced! Check the usage with 'source /path/to/vsc-venv --help'
```

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,252 @@
...
+  local venv_location n_loaded_modules python_version
+
+  requirements_file="$1"
+  modules_file="$2"
+
+  venv_location=$(realpath -m "venvs/venv-${VSC_OS_LOCAL}-${VSC_ARCH_LOCAL}") # full path of venv
+
+  # === Check if requirements file exists === #
+  if [ ! -f "$requirements_file" ]; then
+    echo_error "Requirements file '$requirements_file' not found"
```
boegel, line 93: mention `--requirements` here:

```suggestion
    echo_error "Requirements file '$requirements_file' not found, use --requirements or -r to specify correct path"

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 31]
```diff
@@ -0,0 +1,252 @@
...
+    return 1
+  fi
+
+  echo_info "Requirements installed successfully"
+  echo_info "You can now run your Python scripts"
+}
+
+deactivate_() {
+  echo_info "Deactivating virtual environment at $VIRTUAL_ENV"
+  deactivate # For now, just use the python `deactivate`
```
boegel, line 222: Also run `module purge` here?

boegel (changes requested):
@lbarraga The current contents of the `README` should be moved into a subsection at the end of `setting_up_python_virtual_environments.md`, along with a note/tip on top of that page to recommend using the `vsc-venv` tool for managing/using Python virtual environments of the HPC-UGent clusters.

There should still be a minimal `README` here, which:
i) has a high-level description (one or two sentences on what this tool does);
ii) points to that page (using the full URL: https://docs.hpc.ugent.be/setting_up_python_virtual_environments) for usage info;
iii) says something about "This tool is installed on the HPC-UGent clusters through a module via, ask Kenneth Hoste for more info if needed."

scripts/vsc-venv-wrapper/vsc-venv.sh
```diff
@@ -0,0 +1,252 @@
...
+fi
+
+while [ $# -gt 0 ] ; do
+  case $1 in
+    -a | --activate)      ACTION="activate" ;;
+    -d | --deactivate)    ACTION="deactivate" ;;
+    -r | --requirements)  REQUIREMENTS="$2" ;;
+    -m | --modules)       MODULES="$2" ;;
+    -h | --help)          usage; return 1 ;;
+    -v | --version)       echo "$SCRIPT_NAME v1.0"; return 1;;
```
boegel, line 243: Make a `VERSION` constant on top of this script, and use `${VERSION}` here. No need for non-zero exit when using `--help` or `--version`:

```suggestion
    -h | --help)          usage; return 0 ;;
    -v | --version)       echo "$(basename $SCRIPT_NAME) v${VERSION}"; return 0;;
```

boegel (changes requested):

scripts/vsc-venv-wrapper/vsc-venv.sh
lbarraga: done in [purge modules when deactivating](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/1d896faeee03f0f0ed02b5a68c88a4571e88dbf1)

scripts/vsc-venv-wrapper/README.md
lbarraga: Done in [use 'module load vsc-env'](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/66e03c36b62b6c48335af3322ee07ad89c848855), except when deactivate is called, as the module is already loaded at that time

lbarraga: forgot to [remove .sh](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/241c3dc0698912404434e90a4294cb45d3ac83b7)

lbarraga: is it right that the module is not called `vsc-venv` but `vsc-env`?

lbarraga:
> add README content to python virtual environments docs

done in [add README content to python virtual environments docs](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/2c230b3e83c2e3efc3781651736c9e44c9313b27) and [condense README](https://github.com/hpcugent/vsc_user_docs/pull/670/commits/b49ce7a16b7a6725aef2f2fc2b9fb3e24478b181)

mkdocs/docs/HPC/setting_up_python_virtual_environments.md
```diff
@@ -229,6 +237,106 @@ Two files will be created in the directory where the job was submitted:
...
+## vsc-venv: Python Virtual Environment Wrapper Script
+
+`vsc-venv` is a script that encapsulates the creation and management of Python virtual environments. 
+This avoids multiple issues with the default `venv` included in Python (`python -m venv`).
+
+
+One issue is that a virtual environment created for one cluster might not work on another. 
+Additionally, when activating a virtual environment, 
+the same centrally installed modules used during its creation must also be loaded.
+the `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way,
```
boegel, line 249:
```suggestion
The `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way,

File: mkdocs/docs/HPC/setting_up_python_virtual_environments.md [hunk 34]
```diff
@@ -229,6 +237,106 @@ Two files will be created in the directory where the job was submitted:
...
+
+A virtual environment can be activated by running the following command:
+
+```bash
+$ module load vsc-env
+$ source vsc-venv --activate --requirements <requirements> [--modules module_file]
+```
+
+Here, `requirements` is the path to a file containing the python dependencies to install in the virtual environment.
+An optional `module_file` can be provided, which lists the modules to load before activating the virtual environment.
```
boegel, line 262:
````suggestion
$ source vsc-venv --activate --requirements requirements.txt [--modules modules.txt]
```

Here, `requirements.txt` is the path to a file containing the Python dependencies to install in the virtual environment.
The optional `--modules` option can be used to provide a `modules.txt` file that lists the modules to load before activating the virtual environment.
````

File: mkdocs/docs/HPC/setting_up_python_virtual_environments.md [hunk 35]
```diff
@@ -229,6 +237,106 @@ Two files will be created in the directory where the job was submitted:
...
+
+Now, Python 3.12 is loaded and both the `numpy` (provided by the `SciPy-bundle` module) and `requests` Python packages can be used.
+
+To deactivate the virtual environment, run:
+
+```bash
+source vsc-venv --deactivate
+```
+
+#### making a virtual environment for another cluster
```
boegel, line 319:
```suggestion
#### Making a virtual environment for another cluster

File: mkdocs/docs/HPC/setting_up_python_virtual_environments.md [hunk 36]
```diff
@@ -229,6 +237,106 @@ Two files will be created in the directory where the job was submitted:
...
+
+If we want to create a virtual environment for another cluster, say donphan, we can use `vsc-venv` in an interactive job:
+
+```bash
+$ module swap cluster/donphan
+$ module load vsc-env
+$ qsub -I
+$ cd my_project
+$ source vsc-venv --activate --requirements requirements.txt --modules modules.txt
```
boegel, line 328:
```suggestion
$ qsub -I
$ cd my_project
$ module load vsc-env
$ source vsc-venv --activate --requirements requirements.txt --modules modules.txt
```

scripts/vsc-venv-wrapper/README.md
```diff
@@ -10,86 +10,5 @@ the same centrally installed modules used during its creation must also be loade
 the `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way,
```
boegel, line 10:
```suggestion
The `vsc-venv` command manages multiple virtual environments for different clusters in a transparent way,

**Review by boegel — CHANGES_REQUESTED** [review]
minor tweaks/fixes for docs

File: scripts/vsc-venv-wrapper/vsc-venv.sh [hunk 38]
```diff
@@ -0,0 +1,255 @@
...
+  echo_info "Deactivating virtual environment at $VIRTUAL_ENV"
+  deactivate # For now, just use the python `deactivate`
+  echo_info "Purging loaded modules"
+  module purge
+}
+
+# ============================ Main ============================
+
+SCRIPT_NAME=${BASH_SOURCE[0]} # $0 cannot be used as it gives '-bash' when sourced
+VERSION="1.0"
```
boegel, line 230:
```suggestion
VERSION="1.0.0"

boegel (changes requested):
use `1.0.0` as initial version for `vsc-venv`

lbarraga:
1.0.5 Fixes the following issues:
- because the script is sourced, `MODULES` and `REQUIREMENTS` are loaded into the user shell. When the user runs the program with `--modules modules.txt` and the following run does not contain `--modules modules.txt`, the `MODULES` variable is still set to `modules.txt`.
- fixed an echo that still refered to `--modules` as "the second argument"
- When cluster and host are incompatible, the script suggested to swap and then qsub, but swapping is unnecessary because cluster is already loaded
- with `--help` and `--version`, the script may or may not be sourced. Therefore it is not possible to use a top-level return or exit. Moved invalid action to function which fixes this problem

boegel (approved):

PR #689, opened 2024-09-05
Add Python section to docs

lbarraga:
This PR add a new section to the docs under "Software-Specific Best Practices". It explains how to search and load python packages as modules. It also shows an example on how to run a python script, both interactively and via a jobscript.

### DIFF
--- a/mkdocs/docs/HPC/python.md
+++ b/mkdocs/docs/HPC/python.md
@@ -26,6 +26,18 @@ $ python --version
 Python 3.11.5
 ```
 
+!!! tip
+    To clear all loaded modules, you can run:
+    
+    ```bash
+    $ module purge
+    ```
+
+However, we recommend loading a **Python bundle** module, 
+which already includes a number of commonly used Python packages. 
+Run `module avail Python-bundle-PyPI` to see the available Python bundle modules 
+and `module show <bundle_module>` to see which packages are included in a bundle.
+
 ## Python packages as modules
 
 Python packages can be divided into two categories: standard library packages and third-party packages.
--- a/mkdocs/docs/HPC/python.md
+++ b/mkdocs/docs/HPC/python.md
@@ -190,3 +190,7 @@ After some time, two files will be created in the directory where the job was su
 `python_job_example.o{{jobid}}` and `python_job_example.e{{jobid}}`, where {{jobid}} is the id of your job.
 The `.o` file contains the output of the job.
 
+## More information
+
+- For more info on Python virtual environments, see the [Setting up Python virtual environments](./setting_up_python_virtual_environments.md) page.
+- To use Jupyter on the HPC systems, see the [Jupyter](./jupyter.md) page.
--- a/mkdocs/docs/HPC/python.md
+++ b/mkdocs/docs/HPC/python.md
@@ -159,6 +159,9 @@ Now we can run the script:
 $ python script.py
 ```
 
+For more information on how to start an interactive session, 
+see the [Interactive jobs](./running_interactive_jobs.md) page.
+
 ### Job script
 
 To run the script using a job script, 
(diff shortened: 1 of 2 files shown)


mkdocs/docs/HPC/python.md
```diff
@@ -0,0 +1,164 @@
...
+## Python modules
+
+The HPC systems provide a number of Python modules that you can load into your environment.
+To see a list of available Python modules, you can run:
+
+```bash
+$ module avail Python/
+```
+
+For example, to load Python 3.11, you can run:
```
boegel, line 16: There's also `Python/3.11.5-GCCcore-13.2.0`, maybe use that (since why would you use 3.11.3 if 3.11.5 is there too).

For recent `Python` modules, there are also companion modules that provide a bunch of Python packages from PyPI, like `Python-bundle-PyPI/2023.10-GCCcore-13.2.0`, `SciPy-bundle/2023.11-gfbf-2023b`, `matplotlib/3.8.2-gfbf-2023b`, etc.

These are quite common, so it's worth name-dropping them and mentioning `module show` to inspect what's provided by them so people don't feel a need to install common packages like `matplotlib`, `numpy`, etc.

I would also mention `pip list` to run after loading a bunch of modules, so people can easily check what's readily available already.

mkdocs/docs/HPC/python.md
```diff
@@ -0,0 +1,164 @@
...
+
+```bash
+$ module load Python/3.11.3-GCCcore-12.3.0
```
boegel, line 19: I wouldn't recommend loading this directly, since this is deliberately kept minimal w.r.t. including Python packages (otherwise maintaining these modules is a small nightmare), I would recommend loading `Python-bundle-PyPI` instead, except for older Python versions (< Python 3.11, or modules installed with a toolchain older than `GCC*/12.3.0` or `*/2023a`). With the switch to RHEL9, we'll only have Python 3.11 and newer available via modules, so little room for mistakes when recommending `Python-bundle-PyPI`

mkdocs/docs/HPC/python.md
```diff
@@ -0,0 +1,164 @@
...
+
+To switch to the `donphan` cluster, we run:
+
+```bash
+$ module swap cluster/donphan
+```
+
+### Interactive session
+
+To run the script interactively, we need to start an interactive session on the cluster (`donphan` in this case):
```
boegel, line 119: Like to page/section on interactive jobs?

mkdocs/docs/HPC/python.md
```diff
@@ -0,0 +1,164 @@
...
+To submit the job, run:
+
+```bash
+qsub jobscript.pbs
+```
+
+After some time, two files will be created in the directory where the job was submitted: 
+`python_job_example.o{{jobid}}` and `python_job_example.e{{jobid}}`, where {{jobid}} is the id of your job.
+The `.o` file contains the output of the job.
+
```
boegel, line 164: Add small section/notes at the end to point to dedicated page on i) Python virtual environments ii) Jupyter notebooks

boegel (changes requested):

mkdocs/docs/HPC/python.md
lbarraga: fix in [change python version from 3.11.3 to 3.11.5](https://github.com/hpcugent/vsc_user_docs/pull/689/commits/5b496b381ed62724737ac9b88d1da8a3afe77cf2)

mkdocs/docs/HPC/python.md
lbarraga: fix in [recommend using python bundles over standalone python modules](https://github.com/hpcugent/vsc_user_docs/pull/689/commits/9309c873d4507ddfa238015ba791c1151be817d7)

mkdocs/docs/HPC/python.md
lbarraga: done in [link to interactive sessions page](https://github.com/hpcugent/vsc_user_docs/pull/689/commits/6f3ee91c57d4cec22783f7afdecde0fd8b11de47)

boegel (approved):

PR #690, opened 2024-09-05
add warning to FAQ about PYTHONPATH

lbarraga:


mkdocs/docs/HPC/FAQ.md
```diff
@@ -300,6 +300,11 @@ Note that it is still preferred to submit a software installation request,
 as the software installed by the HPC team will be optimized for the HPC environment.
 This can lead to dramatic performance improvements.
 
+!!! warning 
+    Don't install software using `pip install --user` or `pip install --target` with `$PYTHONPATH`
```
boegel, line 304: `--user` implies installing to `~/.local/lib/python*/site-packages` (maybe mention that)

`$PYTHONPATH` is relevant when stuff is installed with `pip install --prefix` (or maybe `--target`, but that's more uncommon in my experience)

What should also be clearly mentioned is that `$PYTHONPATH` is a "global" setting, independent of the Python versions being used.

Installing Python packages with Python 3.6 with `pip install --prefix` and then updating `$PYTHONPATH` to include `/example/lib/python3.6/site-packages` means trouble when using another Python version for which those Python package installations don't work (since they're specific to Python 3.6).

So, needs a bit more info, I think...

boegel (changes requested):

boegel (approved):

PR #692, opened 2024-09-06
add link that does not work

lbarraga:
I am testing if adding a link to a non existent markdown file makes the CI build fail


lbarraga:
The CI checks failed, which is good.

PR #701, opened 2024-09-10
Auto Update Modules [10 September 2024, 18:18]

hpcugentbot:

Changes:
- Updated the markdown files of all previously available modules
- Added 0 new modules
- Removed 0 modules


boegel (approved):
The removed software versions actually make sense here, the previous update(#699) was created in my VSC account, where additional modules are installed outside of `/apps/gent`...

PR #707, opened 2024-09-12
html detection workflow

lbarraga:
This workflow detects the following simple pattern: `</...>`. It will be used to detect HTML in the codebase.

I chose to keep the pattern simple because matching multi-line strings is more error-prone. Also, links that follow the `<http...>` pattern must not be matched.

This PR also removes the detected HTML tags to make the workflow pass.


lbarraga:
@boegel Ready for review

lbarraga:
HTML has been removed. The CI checks now pass.

boegel (approved):

PR #708, opened 2024-09-12
Remove fetch depth 0 in build workflow

lbarraga:
In the Github action, `fetch-depth: 0` clones the repo with its entire history. I skimmed through the build scripts and it seems we dont do anything with previous commits. Removing the `fetch-depth: 0` makes the workflow only check out the last commit. This brings the "checkout time" form ~30s to ~1s:

`fetch-depth: 0`:
fd0
`fetch-depth: 1 (default)`:
fd1

which brings the total time from ~3m to ~2m30s

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,8 +10,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0 # test
 
       - name: set up Python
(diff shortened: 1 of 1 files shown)


boegel:
@lbarraga That's there for a reason I think. Why do you think we should get rid of it?

boegel:
Hah, I bet I thought that `fetch-depth: 0` does what actually happens by default...

boegel (approved):

PR #742, opened 2024-10-01
skitty upgrade complete

wdpypere:


wdpypere:
link test fails on https://wiki.openfoam.com/Tutorials
I checked manually and the error is: "Resource Limit Is Reached" so bogus

PR #746, opened 2024-10-03
Auto Update Modules [3 October 2024, 15:43]

hpcugentbot:

Changes:
- Updated 6 modules
- Added 0 new modules
- Removed 74 modules


boegel:
A bunch of pages are being removed here because of the fix implemented in #745, which results in cleaning up software-specific pages that shouldn't have been there at all (the changes in #699 were the result of me running the script to auto-update the software overview manually in my VSC account, which resulted in picking up software installations that are not actually there centrally)

PR #780, opened 2024-11-14
add gallade update date

wdpypere:


boegel (approved):
Mon 9 Dec is quite ambitious w.r.t. getting the central software installations ready, but should be doable, so merging...

PR #893, opened 2025-03-12
add Xming website to whitelist, currently broken

boegel:


scripts/link_checker/whitelist.txt
```diff
@@ -51,3 +51,6 @@ https://www.tldp.org/guides.html
 
 # 403
 https://www.edx.org/course/introduction-linux-linuxfoundationx-lfs101x-0
+
+# gone?
+http://www.straightrunning.com/XmingNotes
```
wdpypere, line 56: tests still fail but on `http://www.straightrunning.com/XmingNotes/` with a trailing slash. is it that strict?

PR #1002, opened 2025-08-27
 Change http to https where needed

Oa8choo2:
Change http to https where needed


boegel:
@Oa8choo2 CI issues are bypassed via 883b347012 (which I've included in PR #1005)

PR #1451, opened 2026-08-05
Auto Update Modules [5 August 2026, 07:55]

hpcugentbot:

Changes:
- Updated 566 modules
- Added 5 new modules
- Removed 0 modules


boegel (approved):
this suggests a lot of modules were removed for `skiddo`, but that's not actually the case

it seems like the update of the software overview that was done on 5 July (#1410) was incorrect somehow, modules like `AMICA` or `ABAQUS` versions older than `2025` were never installed for `skiddo` up to now...

